PR #4, opened 2023-05-30
Rely on tsconfig.json rather the command line arguments

### LINKED ISSUE #3: Build ES6 module
IIUC our default package.json commands overwrite the tsconfig.json in ways we don't want.

Currently I think we produce a CommonJS output because the target is not specified in the package .json build command.

I think it would be better to rely on the tsconfig.json.

danyill:
Closes #3

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "test:watch": "tsc && concurrently -k -r \"tsc --watch --preserveWatchOutput\" \"wtr --watch --coverage\" ",
     "doc": "npx rimraf doc && typedoc --out doc index.ts",
     "deploy": "npm run doc && gh-pages --dist 'doc'",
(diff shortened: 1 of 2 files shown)


JakobVogelsang (commented):
I do not think that this PR is really needed anymore. Sorry for the late push of my newest changes. 

I have solved it a bit differently: Use the `esbuildPlugin` from web test runner to build files for tests. Then the tsconfig.ts could be used exclusively for the build. I did it because there is no very nice way to have two tsconfig files in the project and I feel this is necessary because for test we need `include: [**/*.ts]` and for build we rather need `files: [index.ts]`. 

This has a bit of a disadvantage: there are two paths for test and build. Ideally, there is no difference and your tests are close to what you ship. That would then either require to have two tsconfig files and that felt very hecky to do.

danyill:
Thanks for the explanation :-)

PR #13, opened 2023-05-30
feat(tExtRef/extRefTypeRestrictions): get CDC/bType for given pDO/pDA

### LINKED ISSUE #8: (tExtRef) Input restriction for ExtRefs
Similar to #7 we should be able to show the restriction for an ExtRef in terms of the basic type and cdc:

Currently we do:

```ts
/**
 * Edition 2 and later SCL files allow to restrict subscription on
 * later binding type inputs (`ExtRef` elements) based on a `CDC` and
 * basic type `bType`.
 * @param extRef - A later binding type input in the sink IED
 * @returns data objects `CDC` and data attribute basic type `bType` or `null`
 */
export function inputRestriction(extRef: Element): {
  cdc: string | null;
  bType: string | null;
} {
  const [pLN, pDO, pDA] = ['pLN', 'pDO', 'pDA'].map(attr =>
    extRef.getAttribute(attr)
  );
  if (!pLN || !pDO || !pDA) return { cdc: null, bType: null };

  const anyLns = Array.from(
    extRef.ownerDocument.querySelectorAll(
      `LN[lnClass="${pLN}"],LN0[lnClass="${pLN}"]`
    ) ?? []
  );

  for (const anyLn of anyLns) {
    const dataSpec = dataAttributeSpecification(anyLn, pDO, pDA);
    if (dataSpec.cdc !== null && dataSpec.bType !== null) return dataSpec;
  }

  return { cdc: null, bType: null };
}

```

I'd like to export this so I can show this in a hover for ExtRefs in the later binding plugin.

There is a little question in my mind around the way pDO is used.

Currently we use the pDO to derive the CDC but Ed 2.1 appears to also allow the option of a DO _name_.  Either this should be a parameter or it is just a mistake in the standard:

image

Most of part 6 and part 7-1 suggest that the pDO should only be used for the CDC but there is this inconsistency

JakobVogelsang:
…n pDO/pDA

Closes #8 

As discussed in, the issue is based on the IEC 61850 namespace instead of relying on to parse the data model. It can be used as follows

```ts
const {cdc,bType} = extRefTypeRestrictions(extRef);

```


tExtRef/extRefTypeRestrictions.ts
```diff
@@ -0,0 +1,32 @@
+import { dataObjects } from "../foundation/nsd.js";
+
+/**
+ * This function returns the common data class `CDC` of the
+ * specified data object `pDO` and the type of the specified
```
danyill, line 5:
```suggestion
 * specified data object `pDO` and the basic type of the specified
```

danyill (approved):
A few small typos but looks good to me and I'm excited that this will likely provide better performance.

PR #22, opened 2023-06-04
feat(tFCDA): determin CDC and bType

### LINKED ISSUE #7: (tLN) Data attribute specification
For subscriptions and support of pXX types it would be useful to have a function to bring back the common data class and basic type of a data attribute.

This can be also useful to show the basic type and cdc in the UI (I do this in the later binding plugin on hover).

image

We currently do something like this:

```ts
function dataAttributeSpecification(
  anyLn: Element,
  doName: string,
  daName: string
): { cdc: string | null; bType: string | null } {
  const doc = anyLn.ownerDocument;
  const lNodeType = doc.querySelector(
    `LNodeType[id="${anyLn.getAttribute('lnType')}"]`
  );

  const doNames = doName.split('.');
  let leaf: Element | null | undefined = lNodeType;
  for (const doN of doNames) {
    const dO: Element | null | undefined = leaf?.querySelector(
      `DO[name="${doN}"], SDO[name="${doN}"]`
    );
    leaf = doc.querySelector(`DOType[id="${dO?.getAttribute('type')}"]`);
  }
  if (!leaf || !leaf.getAttribute('cdc')) return { cdc: null, bType: null };

  const cdc = leaf.getAttribute('cdc')!;

  const daNames = daName.split('.');
  for (const daN of daNames) {
    const dA: Element | null | undefined = leaf?.querySelector(
      `DA[name="${daN}"], BDA[name="${daN}"]`
    );
    leaf =
      daNames.indexOf(daN) < daNames.length - 1
        ? doc.querySelector(`DAType[id="${dA?.getAttribute('type')}"]`)
        : dA;
  }
  if (!leaf || !leaf.getAttribute('bType')) return { cdc, bType: null };

  const bType = leaf.getAttribute('bType')!;

  return { bType, cdc };
}
```

Only question being should we make this composable so that a cdc can be brought back separately for a data object? I think the answer is probably.

JakobVogelsang:
Closes #7

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -13,3 +13,5 @@ export { lnInstGenerator } from "./generator/lnInstGenerator.js";
 export { UnsubscribeOptions } from "./tExtRef/unsubscribe.js";
 export { unsubscribe } from "./tExtRef/unsubscribe.js";
 export { extRefTypeRestrictions } from "./tExtRef/extRefTypeRestrictions.js";
+
+export { fcdaBaseTypes } from "./tFCDA/fcdaBaseTypes.js";
--- a/tFCDA/fcdaBaseTypes.spec.ts
+++ b/tFCDA/fcdaBaseTypes.spec.ts
@@ -1,0 +1,68 @@
+import { expect } from "chai";
+
+import { findElement } from "../foundation/helpers.test.js";
+import { publisherIED } from "./fcdaBaseTypes.testfiles.js";
+
+import { fcdaBaseTypes } from "./fcdaBaseTypes.js";
+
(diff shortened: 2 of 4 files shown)


danyill (changes requested):
I've made a few small naming and selector suggestions.

One of the performance limitations of the subscriber later binding plugin is that it needs to calculate the `CDC` and `bType` for every FCDA to know whether it should be shown as disabled when I click on an ExtRef in the subscriber view.

This is costly and may potentially be a candidate for some kind of indexing service in OpenSCD. I guess it could be sped up by grouping the hierarchy but the queries are already quite specific.

I really like this very tidy triplet of files for a function - very pleasing to look at.

JakobVogelsang:
All very valid points! Thank you. I shall change that.

PR #24, opened 2023-06-08
feat(tExtRef/subscribe): initial version

### LINKED ISSUE #9: (tExtRef) API for subscription
Jakob asked:

>  The simple API for such a subscribe  function seems to be impossible, if you take the historical changes and the many used cases subscribing is covering. I therefore have come up with something like this:
> 
> ```ts
> function subscribe(options: SubscriberOption[]): (Update | Insert) []
> ```
> 
> where
> 
> ```ts
> SubscriberOption = {sink: Element, source: {fcda: Element, controlBlock?:Element} | Source}
> ```
> 
> and
> 
> ```ts
> Source = {
>  sourceIed: Element;
>  iedName: string;
>  ldInst: string;
>  prefix: string | null;
>  lnClass: string;
>  lnInst: string | null;
>  doName: string;
>  daName: string | null;
>  fc: string;
>  serviceType?: 'Report' | 'GOOSE' | 'SMV' | 'Poll' | null;
>  srcCBName?: string;
>  srcLDInst?: string;
>  srcPrefix?: string | null;
>  srcLNClass?: string;
>  srcLNInst?: string | null;
> }
> ```
> 
> For now transformer is from `{fcda: Element, controlBlock?: Element}` to Source and internally work with this Object and after failing to cover all scenarious I am thinking of allowing to set source through Source and therefore allow a flexible definition. 
> 
> For the user it does mean more transformation work, but there is still good checks and the automated subscription supervision handling in the background.
> 
> There are two cases where fcda is not sufficient. The main used case is specification where there is only data attributes bind to external reference and we also need to have something to allow bind Poll kind of data where an FCDA as the source is also not sufficient enough.
> 
> The main reason I input an SubscriptionOption instead of just three elements is that we need to allow an array of subscription at once. Doing them one after the other and just fire several events would lead to wrong subscription supervision insert. Those would also lead to incorrect checks of already existing ExtRefs and other phenomenos of that sort. We check agains the SCL and with multiple single events there would be an invalid in.between of the SCL.

Thinking out loud a bit and perhaps being a bit shallow? :stuck_out_tongue_winking_eye: 

I think that the propose SubscriptionOption is too complex and suggest we should have separate functions with simpler, end-user oriented call signatures and APIs to ease adoption and educate the user into the "typical possibilities for actions".

How about:

```ts
function subscribe(subscriptionObject[], options?: Options): (Update | Insert) []
```

```ts
type subscriptionObject = {
  extRef: Element
  fcd: Element,
};
```

Here `fcd` could be a DA or a DO?

```ts
type Options = {
  ignoreSupervisions: boolean,
  isPoll: boolean
  isSpecification: boolean
};
```

I think for other requirements we should have different functions or end-user oriented options, perhaps broadly aligned to Table 51 of -6 Ed 2.1, e.g.

* Adapting pTypes (if that is a valid use case for an SCT)
* If using `isPoll` then an option can allow ignoring of the source control block. 
* If carrying out a specification,  `isSpecification` as an option, then the FCDA is "constructed" by the user (perhaps with a separate function) because the FCDA attributes and the ExtRef attributes are identical for this use case I think.

Looking forward to some discussion.

JakobVogelsang:
Closes #9 with an initial `subscribe` function that supports:

- check for valid source and sink definitions
- later binding
- srcXXX attributes
- check for ExtRef restrictions

and does not:
- add subscription supervision
- physical connection

### DIFF
--- a/foundation/helpers.test.ts
+++ b/foundation/helpers.test.ts
@@ -1,4 +1,9 @@
-export function findElement(str: string, selector: string): Element | null {
+export function findElement(
+  str: string,
+  selector?: string
+): XMLDocument | Element | null {
+  if (!selector) return new DOMParser().parseFromString(str, "application/xml");
+
   return new DOMParser()
     .parseFromString(str, "application/xml")
     .querySelector(selector);
--- a/foundation/utils.ts
+++ b/foundation/utils.ts
@@ -27,6 +27,9 @@ export function isRemove(edit: Edit): edit is Remove {
     (edit as Insert).parent === undefined && (edit as Remove).node !== undefined
   );
 }
+export function isInsert(edit: Edit): edit is Insert {
+  return (edit as Insert).parent !== undefined;
+}
 
 /** Utility function to create element with `tagName` and its`attributes` */
 export function createElement(
--- a/index.ts
+++ b/index.ts
@@ -13,6 +13,7 @@ export { lnInstGenerator } from "./generator/lnInstGenerator.js";
 export { UnsubscribeOptions } from "./tExtRef/unsubscribe.js";
 export { unsubscribe } from "./tExtRef/unsubscribe.js";
 export { extRefTypeRestrictions } from "./tExtRef/extRefTypeRestrictions.js";
+export { doesFcdaMeetExtRefRestrictions } from "./tExtRef/doesFcdaMeetExtRefRestrictions.js";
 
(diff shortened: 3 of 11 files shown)


danyill:
Sorry this is taking me a little while. I hope to look at this PR in suitable detail this week. Thank you for the work on this key feature.

PR #29, opened 2023-06-14
Ensure serviceType handled correctly on unsubscribe, closes #27

### LINKED ISSUE #27: tExtRef: Unsubscribe doesn't handle serviceType correctly
What we do during an unsubscribe is set to the `serviceType` to the value of this function:

https://github.com/OpenEnergyTools/scl-lib/blob/cc81a2341aedcb44b469013aacedf4a26c29c438/tExtRef/unsubscribe.ts#L14-L21

<s>However we should be setting to null rather than undefined I think as we do for other attributes.</s> NO!

I think we might want something like:

```js
updateActions.push({
                element: extRef,
                attributes: {
                    iedName: null,
                    ldInst: null,
                    prefix: null,
                    lnClass: null,
                    lnInst: null,
                    doName: null,
                    daName: null,
                    srcLDInst: null,
                    srcPrefix: null,
                    srcLNClass: null,
                    srcLNInst: null,
                    srcCBName: null,
                    ...(extRef.getAttribute("pServT") && {serviceType: null})
                },
            });
```

image

As it stands, what happens is this causes an error during event handling.

image

danyill:


JakobVogelsang (changes requested):
Thank you for your contribution. I would love to have two little changes:

1. Please adapt the line 67 in that same test file `unsubscribe.spec.ts`
2. Please fixup the two and when added reviews to one commit. I would love to work with rebase on mergin to main and having additional commits would read strage.

danyill:
I have adapted L67:

I have done this:

```ts
expect(update.attributes).to.not.have.own.property("serviceType");
```

To me, checking that it is undefined also seems fine, I just like that this is more explicit.
Perhaps I have not understood your comment.


>Please fixup the two and when added reviews to one commit. I would love to work with rebase on mergin to main and having additional commits would read strage.

Sure, sounds fun.

JakobVogelsang (approved):

PR #30, opened 2023-06-22
fix(foundation/nsd): Improve nsd import in dependent libraries

### LINKED ISSUE #28: Error when using this library in another project
I can't integrate the current version of scl-lib in my project due to MIME typing checking. I receive the following console error when using scl-lib version 0.1.0 for `nsd.json`

> Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "application/json". Strict MIME type checking is enforced for module scripts per HTML spec.

image

danyill:
Closes #28


danyill:
One concern with this approach is performance. The `nsd.ts` is 2.4 Mb.

OTOH it compresses via zip down to 31.7 KiB so transfer time on a client<>server connection with a modern web server should be negligible and I assume that because JSON is a native format it can be efficiently used.

I'm curious about the exact area that performance could be an issue. If I wrap the import of the typescript in a timer, it appears to only be a few ms to load.

ca-d:
I don't currently see how an `import data from 'something.js'` statement should be significantly more or less performant than a top level `await fetch('something.json').then(res => res.json())`. @JakobVogelsang in what scenario do you fear a performance degradation?

In general, I'd say that unless an inferred type is really needed for this data, fetching a `json` file at load time is the cleaner solution, since if we're loading only data and don't need to execute any code, a pure data format is the simpler way to go.

danyill:
We will rework and try a cleaner solution, thanks for taking a look :+1:

PR #31, opened 2023-06-23
fix(foundation/nsd): Improve nsd import in dependent libraries

### LINKED ISSUE #28: Error when using this library in another project

danyill:
Closes #28

This almost works but I can't get the tests to pass:

```
tExtRef/extRefTypeRestrictions.spec.ts:

 🚧 Browser logs:
      SyntaxError: Unexpected token 'v', "var NamPlt"... is not valid JSON

 ❌ Could not import your test module. Check the browser logs or open the browser in debug mode for more information. 
```

I think this is because "somehow" a sourcemap is included with this json file which I must still examine.

### DIFF
--- a/tExtRef/extRefTypeRestrictions.ts
+++ b/tExtRef/extRefTypeRestrictions.ts
@@ -1,4 +1,6 @@
-import dataObjects from "../foundation/nsd.json";
+const dataObjects = await fetch(
+  new URL("../foundation/nsd.json", import.meta.url)
+).then((res) => res.json());
 
 /**
  * This function returns the common data class `CDC` of the
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -17,5 +17,6 @@
   },
   "files": [
     "./index.ts",
+    "./foundation/nsd.json"
   ],
 }
--- a/web-test-runner.config.mjs
+++ b/web-test-runner.config.mjs
@@ -5,7 +5,7 @@ export default /** @type {import("@web/test-runner").TestRunnerConfig} */ ({
   /** Test files to run */
   files: '**/*.spec.ts',
 
-  plugins: [esbuildPlugin({ ts: true, json: true })],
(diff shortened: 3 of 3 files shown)


tExtRef/extRefTypeRestrictions.ts
```diff
@@ -1,4 +1,16 @@
...
+    throw new Error(
+      `Failed to fetch JSON file: ${path.toString()} ${response.status} ${
+        response.statusText
+      }`
+    );
+  }
+  return await response.json();
+}
+
+const dataObjects = await fetchJSON("../foundation/nsd.json");
```
ca-d, line 13: What's the functional difference between this and

```suggestion
const dataObjects = await fetch(new URL("../foundation/nsd.json", import.meta.url)).then(res => res.json());
```

apart from the custom error message? I'm asking quite naively, I assume that both would just result in a rejected promise in case of either a non-ok response or invalid json, but I haven't tried. If and only if there is a significant functional advantage I would keep the custom function, otherwise I'd probably prefer the one-liner for readability.

ca-d (approved):
LGTM, great detective work figuring out how to wrangle the impertinent build tool! :mag:

tExtRef/extRefTypeRestrictions.ts
danyill: I agree, let's do that.

danyill:
Thank you for reviewing. I have squashed and added the suggestion (as Jakob would like to use rebase on this project rather than merge)

JakobVogelsang (approved):
LGMT

I have learned as well a bit more about the fetch API

PR #33, opened 2023-06-26
Subscribe

### LINKED ISSUE #9: (tExtRef) API for subscription

JakobVogelsang:
Closes #9

### DIFF
--- a/tExtRef/subscribe.ts
+++ b/tExtRef/subscribe.ts
@@ -1,0 +1,264 @@
+import { Insert, Update, createElement } from "../foundation/utils.js";
+import { getReference } from "../general/getReference.js";
+
+import { doesFcdaMeetExtRefRestrictions } from "./doesFcdaMeetExtRefRestrictions.js";
+
+const serviceTypes: Partial<Record<string, "Report" | "GOOSE" | "SMV">> = {
+  ReportControl: "Report",
+  GSEControl: "GOOSE",
+  SampledValueControl: "SMV",
+};
+
+export type SubscribeOptions = {
+  /** Disables checks before subscription. !!Use with care */
+  force: boolean;
+};
+
+export type Connection = {
+  /** Can be `LN0`, `LN`, `Inputs` or `ExtRef` */
+  sink: Element;
+  source: {
+    /** `FCDA` element with or without `daName` */
+    fcda: Element;
+    /** `ReportControl`, `GSEControl`, `SampledValueControl` */
+    controlBlock?: Element;
+  };
+};
+
+function srcAttributes(controlBlock?: Element): {
+  srcLDInst: string | null;
+  srcPrefix: string | null;
+  srcLNClass: string | null;
+  srcLNInst: string | null;
+  srcCBName: string | null;
+  serviceType: "Report" | "GOOSE" | "SMV" | null;
+} | null {
+  const srcLDInst =
+    controlBlock?.closest("LDevice")?.getAttribute("inst") || null;
+  const srcPrefix =
+    controlBlock?.closest("LN0,LN")?.getAttribute("prefix") || "";
+  const srcLNClass =
+    controlBlock?.closest("LN0,LN")?.getAttribute("lnClass") || null;
+  const srcLNInst =
+    controlBlock?.closest("LN0,LN")?.getAttribute("inst") || null;
+  const srcCBName = controlBlock?.getAttribute("name") || null;
+
+  if (
+    !controlBlock ||
+    !serviceTypes[controlBlock.tagName] ||
+    !srcLDInst ||
+    !srcLNClass ||
+    !srcCBName
+  )
+    return {
+      srcLDInst: null,
+      srcPrefix: null,
+      srcLNClass: null,
+      srcLNInst: null,
+      srcCBName: null,
+      serviceType: null,
+    };
+
+  return {
+    srcLDInst,
+    srcPrefix,
+    srcLNClass,
+    srcLNInst,
+    srcCBName,
+    serviceType: serviceTypes[controlBlock.tagName]!,
+  };
+}
+
+function getDataAttributes(fcda: Element): {
+  iedName: string;
+  ldInst: string;
+  prefix: string | null;
+  lnClass: string;
+  lnInst: string | null;
+  doName: string;
+  daName: string | null;
+  fc: string;
+} | null {
+  const sourceIed = fcda.closest("IED");
+  const iedName = sourceIed?.getAttribute("name");
+  const [ldInst, prefix, lnClass, lnInst, doName, daName, fc] = [
+    "ldInst",
+    "prefix",
+    "lnClass",
+    "lnInst",
+    "doName",
+    "daName",
+    "fc",
+  ].map((attr) => fcda.getAttribute(attr));
+
+  if (!sourceIed || !iedName || !ldInst || !lnClass || !doName || !fc)
+    return null;
+
+  return {
+    iedName,
+    ldInst,
+    prefix,
+    lnClass,
+    lnInst,
+    doName,
+    daName,
+    fc,
+  };
+}
+
+function createSubscribeAction(
+  connection: Connection,
+  parent: Element
+): Update | Insert | null {
+  const doc = connection.sink.ownerDocument;
+  const fcda = connection.source.fcda;
+  const controlBlock = connection.source.controlBlock;
+
+  const isEd1 = !doc.querySelector("SCL")?.getAttribute("version");
+
+  const ed1Attributes = getDataAttributes(fcda);
+  if (!ed1Attributes) return null;
+
+  const ed2Attributes = {
+    ...ed1Attributes,
+    ...srcAttributes(controlBlock),
+  };
+
+  if (connection.sink.tagName === "ExtRef" && isEd1)
+    return {
+      element: connection.sink,
+      attributes: ed1Attributes,
+    };
+
+  if (connection.sink.tagName === "ExtRef" && !isEd1)
+    return {
+      element: connection.sink,
+      attributes: ed2Attributes,
+    };
+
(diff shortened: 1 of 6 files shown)


danyill (approved):
This is a really nice piece of work and I'm keen to see it merged and a new release!

The code paths are much clearer, simpler and more powerful than our existing subscription functionality.

I will be delighted to include in my oscd-subscriber-later-binding plugin and have tested the functionality with only one comment which is around philosophy of how our functions should work.

I like that it will efficiently allow FCDO subscriptions (not currently supported in my plugin but not difficult to) without much difficulty.

One question around the call signature: How should we handle subscription to a data attribute which is part of an FCDO (i.e. the FCDA does not include the data attribute explicitly)? This could be done in a later iteration but I  worry about backwards compatibility.

tExtRef/subscribe.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  return !(
+    sink.tagName === "LN" ||
+    sink.tagName === "LN0" ||
+    sink.tagName === "Inputs"
+  );
+}
+
+function validSubscribeConditions(connection: Connection): boolean {
+  if (invalidSink(connection.sink)) return false;
```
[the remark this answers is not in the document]
danyill: Maybe a `force` flag?

Ideally we should not force the user to push multiple events for a set of edits IMHO.

tExtRef/subscribe.ts
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * A function to subscribe [[`source`]] to [[`sink`]].
+ * @param sink - Can be `LN0`, `LN` and `Inputs`
+ * for non-later-binding type of subscription and `ExtRef` with `intAddr`
+ * for later binding type subscription.
+ * @param source.fcda - `FCDA` element
+ * @param source.controlBlock - The control block carrying the [[`source.fcda`]]
+ * @returns An array of actions to do a valid subscription
+ */
+export function subscribe(connections: Connection[]): (Insert | Update)[] {
```
danyill, line 227: I was talking to Christian, we thought it would be nice if we could pass a single connection so the call signature didn't require an array.

tExtRef/subscribe.ts
```diff
@@ -0,0 +1,238 @@
...
+  const validConnections = connections.filter(
+    validSubscribeConditions
+  ) as Connection[];
+
+  const extRefActions = createSubscribeActions(validConnections);
+
+  return [
+    ...extRefActions,
+    //TODO: ...insertSubscriptionSupervisions(extRefActions),
```
danyill, line 236:
```suggestion
    //See https://github.com/OpenEnergyTools/scl-lib/issues/26
    //TODO: ...insertSubscriptionSupervisions(extRefActions),
```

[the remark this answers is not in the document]
JakobVogelsang: I have that on the radar. I would like to make it default, so you would have to still touch the code once it is in. Is that ok with you?

tExtRef/subscribe.ts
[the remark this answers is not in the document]
JakobVogelsang: I like the `force` option. I again would love to make it default false, but at least you can then use it without unsubscribing it. That would be strange behavior in the history later on.

tExtRef/subscribe.ts
JakobVogelsang: You mean to avoid calling this function like so: `subscribe([conn])` and be able to call it like so `subscribe(conn)`?

JakobVogelsang:
Regarding the subscription to `FCD`: I myself am not fully sure. This is how I understand it:

You can subscribe an FCD in two ways - with later binding only with missing `pDO` or `pDO` present but missing `pDA`:
1. Have `ExtRef` without `daName` (it is up to the vendor tool to pick the data from the DO with a specific `fc`). This is the way this function is doing it right now. 
2. Subscribe all or some of the data attributes with the defined `fc` individually. Here we would need to extend.

Describing FCD types to later binding type `ExtRef` is a tricky business, though. So far I understood it is up to the ICT tool what of the data attributes is accepted in the end. For documentation purpose, it therefore is necessary to have a roundtrip over the ICT tool. Exactly for this purpose, I have chosen the variant 1. With that, at least you make sure all data attributes are available to the ICT tool at the end.

For variant 2 we would need to have an extra key in the source like so:

```ts
source: {
 fcda: Element;
 dataAttributes?: string[];
 controlBlock?: Element;
}
```

If you want to know how variant 2 would look like, take a look at Annex H Used Case 1.

tExtRef/subscribe.ts
danyill: Yip. Exactly perhaps also for unsubscribe

JakobVogelsang: I am a bit hesitated to do so. My reason for this is this: I think it might be hard for some users of this library to understand that a correct result can only be achieved when right after determining the diff you must immediately send an edit event so that this diff is put into the SCL right away. The reason is that this function is assuming correct SCL file. I am afraid that with the changes you request it might be more inviting to do something like this

```ts
const diff1 = subscribe(conn1);
const diff2 = subscribe(conn2)

dispatch(editEvent([...diff1,...diff2]));
```   

This however gives me the hint to put that fact into the doc with big exclamation marks and a bad example. 

Now, when you are of the opinion that I am overstating this risk, I am willing to take the chances on this :).

danyill: Do you mean if one of these is an `unsubscribe` event? More like this:

```ts
const diff1 = unsubscribe(conn1);
const diff2 = subscribe(conn2)

dispatch(editEvent([...diff1,...diff2]));
```

I don't think that's a reason not to have a simpler or compatible API. It's just there are some limitations around composability. I think we need a warning something like:

> `subscribe` and `unsubscribe` functions return events. If these functions are called on the same ExtRef elements they are not composable within the same `oscd-event` unless the `force` flag is used. Where possible avoid using the force flag and use separate `unsubscribe` events prior to `subscribe` events (to ensure checks for preconfigured types, not overriding existing subscriptions and correct elements occur).

danyill (approved):
`force` flag LGTM

tExtRef/subscribe.ts
JakobVogelsang: I am with you and am convinced. Just to clarify the problem that might occur. Let's iamgine we have a SCL element like so:

```xml
<LN ...>
...
<Inputs>
  <ExtRef ... />
  <ExtRef ... />
</Inputs>
</LN>
```

Let's further assume a user of this library will use the `unsubscribe` function like so:

```ts
const diff1 = unsubscribe([extRef1]);
const diff2 = unsubscribe([extRef2]);

...dispatch(editEvent([diff1,diff2]));
```

will leave an result like so:
```xml
<LN ... >
   <Inputs />
</LN>
```

Compared to this:
```ts
.dispatch(editEvent(ubsubscribe[extRef1,extRef2]))
```

will leave the correct result like so:

```ts
<LN ...>
 ....
</LN>
```

PR #38, opened 2023-07-16
feat(tLN/removeSubscriptionSupervision)

### LINKED ISSUE #6: (tLN) Supervision function selector scoping
The selectors we use in our supervision functions are too broad:

```js
/** @returns Element to remove the subscription supervision */
function removableSupervisionElement(
  ctrlBlock: Element,
  subscriberIed: Element
): Element | null {
  const supervisionType = ctrlBlock.tagName === "GSEControl" ? "LGOS" : "LSVS";

  const valElement = Array.from(
    subscriberIed.querySelectorAll(
      `LN[lnClass="${supervisionType}"] > DOI > DAI > Val`
    )
  ).find((val) => val.textContent === controlBlockObjRef(ctrlBlock));
  if (!valElement) return null;

  const ln = valElement.closest("LN")!;
  const doi = valElement.closest("DOI")!;

  // do not remove logical nodes `LGOS`, `LSVS` unless privately tagged
  const canRemoveLn = ln.querySelector(
    ':scope > Private[type="OpenSCD.create"]'
  );

  return canRemoveLn ? ln : doi;
}
```

This is too broad: `LN[lnClass="${supervisionType}"] > DOI > DAI > Val`

The DAI must be scoped to `setSrcRef` and the DOI must be scoped to `GoCBRef` or `SvCBRef`

I have generally improved these as I've noticed this in a few places: https://github.com/danyill/oscd-subscriber-later-binding/blob/main/foundation/subscription/subscription.ts

I can do a PR to fix this if you agree.

### LINKED ISSUE #18: (tLN) Remove subscription supervision
As a user when I remove a subscription supervision I may want to remove the supervision LN or simply remove the control block reference.

I would like us to have an option for this.

My use case is that I wish to map the supervision instances in an MMS data flow and so as users subscribe and unsubscribe the only thing which changes is the control block reference.

danyill:
* Fix supervision function selector scoping (closes #6)
* Allow only CB or LN removal for subscription supervision removal (closes #18)

<s>WIP, still to do:

* SV tests
* Tests `for canRemoveSubscriptionSupervision`</s>

This breaks the previous API. Subscription removal only either removes the LN or the `textContent` of the `Val` element. OpenSCD created supervision elements are not treated in any special way.
Additionally the first supervision in the data model cannot be removed (and trying only yields removal of the `textContent` of the `Val` element).

<s>I'll complete this soon if this seems like the correct direction @JakobVogelsang </s>

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,3 +1,5 @@
+import { canRemoveSubscriptionSupervision } from "./tLN/canRemoveSubscriptionSupervision.js";
+
 export { Update } from "./foundation/utils.js";
 export { Insert } from "./foundation/utils.js";
 export { Remove } from "./foundation/utils.js";
@@ -17,6 +19,9 @@ export { SubscribeOptions } from "./tExtRef/subscribe.js";
 export { subscribe } from "./tExtRef/subscribe.js";
 export { extRefTypeRestrictions } from "./tExtRef/extRefTypeRestrictions.js";
 export { doesFcdaMeetExtRefRestrictions } from "./tExtRef/doesFcdaMeetExtRefRestrictions.js";
+
+export { removeSubscriptionSupervision } from "./tLN/removeSubscriptionSupervision.js";
+export { canRemoveSubscriptionSupervision } from "./tLN/canRemoveSubscriptionSupervision.js";
 
 export { fcdaBaseTypes } from "./tFCDA/fcdaBaseTypes.js";
 
--- a/tExtRef/unsubscribe.spec.ts
+++ b/tExtRef/unsubscribe.spec.ts
@@ -5,7 +5,14 @@ import {
   withSubscriptionSupervision,
 } from "./unsubscribe.testfiles.js";
 
-import { Remove, Update, isRemove, isUpdate } from "../foundation/utils.js";
+import {
+  Insert,
+  Remove,
+  Update,
+  isInsert,
+  isRemove,
+  isUpdate,
+} from "../foundation/utils.js";
 import { unsubscribe } from "./unsubscribe.js";
 import { findElement } from "../foundation/helpers.test.js";
 
@@ -99,17 +106,25 @@ describe("Function allowing to unsubscribe multiple external references", () => 
       'LN[lnClass="LGOS"][inst="2"]'
     );
 
-    expect(actions.length).to.equal(5);
+    expect(actions.length).to.equal(7);
     expect(actions[0]).to.satisfies(isRemove);
     expect((actions[0] as Remove).node).to.equal(extRefs[0]);
     expect(actions[1]).to.satisfies(isRemove);
     expect((actions[1] as Remove).node).to.equal(extRefs[1]);
     expect(actions[2]).to.satisfies(isUpdate);
     expect((actions[2] as Update).element).to.equal(extRefs[2]);
-    expect(actions[3]).to.satisfies(isRemove);
-    expect((actions[3] as Remove).node).to.equal(doi);
-    expect(actions[4]).to.satisfies(isRemove);
-    expect((actions[4] as Remove).node).to.equal(ln);
+
+    expect(isRemove(actions[3])).to.be.true;
+    expect((<Remove>actions[3]).node.nodeName).to.be.equal("Val");
+    expect((<Remove>actions[3]).node.textContent).to.be.equal(
+      "srcIEDsomeLDInst/LLN0.someGse"
+    );
+
+    expect(isInsert(actions[4])).to.be.true;
+    expect((<Insert>actions[4]).node.nodeName).to.be.equal("Val");
+    expect((<Insert>actions[4]).node.textContent).to.be.equal("");
+    expect((<Insert>actions[4]).parent.nodeName).to.be.equal("DAI");
+    expect((<Insert>actions[4]).reference).to.be.null;
   });
 
   it("with ignoreSupervision do not remove subscription LGOS supervision", () => {
@@ -162,15 +177,26 @@ describe("Function allowing to unsubscribe multiple external references", () => 
     );
     const actions = unsubscribe(extRefs);
 
-    expect(actions.length).to.equal(4);
+    expect(actions.length).to.equal(5);
+
     expect(actions[0]).to.satisfies(isRemove);
     expect((actions[0] as Remove).node).to.equal(extRefs[0]);
     expect(actions[1]).to.satisfies(isRemove);
     expect((actions[1] as Remove).node).to.equal(extRefs[1]);
     expect(actions[2]).to.satisfies(isRemove);
     expect((actions[2] as Remove).node).to.equal(extRefs[1].parentElement);
-    expect(actions[3]).to.satisfies(isRemove);
-    expect((actions[3] as Remove).node).to.equal(doi);
+
+    expect(isRemove(actions[3])).to.be.true;
+    expect((<Remove>actions[3]).node.nodeName).to.be.equal("Val");
+    expect((<Remove>actions[3]).node.textContent).to.be.equal(
+      "srcIEDsomeLDInst/LLN0.someSmv"
+    );
+
+    expect(isInsert(actions[4])).to.be.true;
+    expect((<Insert>actions[4]).node.nodeName).to.be.equal("Val");
+    expect((<Insert>actions[4]).node.textContent).to.be.equal("");
+    expect((<Insert>actions[4]).parent.nodeName).to.be.equal("DAI");
+    expect((<Insert>actions[4]).reference).to.be.null;
   });
 
   it("with ignoreSupervision do not remove subscription LGOS supervision", () => {
--- a/tLN/canRemoveSubscriptionSupervision.spec.ts
+++ b/tLN/canRemoveSubscriptionSupervision.spec.ts
@@ -1,0 +1,123 @@
+import { expect } from "chai";
+
+import {
+  gooseSubscriptionEd2,
+  svSubscriptionEd2,
+} from "./removeSubscription.testfiles.js";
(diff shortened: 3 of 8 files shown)


index.ts
```diff
@@ -18,6 +20,9 @@ export { subscribe } from "./tExtRef/subscribe.js";
 export { extRefTypeRestrictions } from "./tExtRef/extRefTypeRestrictions.js";
 export { doesFcdaMeetExtRefRestrictions } from "./tExtRef/doesFcdaMeetExtRefRestrictions.js";
 
+export { removeSubscriptionSupervision } from "./tLN/removeSubscriptionSupervision.js";
+export { canRemoveSubscriptionSupervision } from "./tLN/canRemoveSubscriptionSupervision.js";
```
danyill, line 24: I guess we'll need to refactor this for adding supervisions with a similar function to check if we can create a supervision instance.

JakobVogelsang: Partly that is true. For the creation of a subscription, supervision I at the moment check for this:
- is SrcRef editable (that is the check we need here too)
- exist yet another `form` LGOS or LSVS that does pin the parent for those logical nodes in the subscriber IED (we do not need this check when we remove)
- whether a control block is already supervised in the subscriber IED (also not neccassary)

tLN/removeSubscriptionSupervision.ts
```diff
@@ -142,21 +63,48 @@ function groupPerControlBlock(
  * @param extRefs - An array of external reference elements
  * @returns Actions to remove subscription supervision `LGOS` or `LSVS`
  */
-export function removeSubscriptionSupervision(extRefs: Element[]): Remove[] {
-  if (extRefs.length === 0) return [];
+export function removeSubscriptionSupervision(
+  extRefOrExtRefs: Element | Element[],
+  options?: { removeLN: boolean }
```
JakobVogelsang, line 68: I would like to add a default options object here. That read better and can be better handled in the function below like so:

`(..., options = {removeLN: false}): (Remove | Insert)[]`

tLN/canRemoveSubscriptionSupervision.ts
```diff
@@ -0,0 +1,109 @@
...
+} from "./removeSubscriptionSupervision";
+
+/** @returns Whether `DA` with name `setSrcRef` can edited by SCL editor */
+function isSrcRefEditable(ctrlBlock: Element, subscriberIed: Element): boolean {
+  const supervisionElement = removableSupervisionElement(
+    ctrlBlock,
+    subscriberIed
+  );
+
+  const ln = supervisionElement?.closest("LN") ?? null;
```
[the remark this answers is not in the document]
JakobVogelsang: I just see that is something I have introduced, sorry for this. But as you are touching it, I would love this to be removed. It reads better I think.

tLN/canRemoveSubscriptionSupervision.ts
```diff
@@ -0,0 +1,109 @@
+import {
+  GroupedExtRefs,
+  groupPerControlBlock,
+  removableSupervisionElement,
+} from "./removeSubscriptionSupervision";
+
+/** @returns Whether `DA` with name `setSrcRef` can edited by SCL editor */
```
JakobVogelsang, line 7: My spelling error I would love you to fix. `...setSrcRef can be edited`. I would in addition make the doc more precise and mention `DAI` as well like so: `Whether `DA` | `DAI`...`

tLN/canRemoveSubscriptionSupervision.ts
```diff
@@ -0,0 +1,109 @@
...
+  );
+}
+
+/**
+ * Checks if a group of ExtRefs can have their supervision references removed
+ * @param extRefOrExtRefs - An SCL ExtRef element or array of them.
+ * @returns true if all supervision references can be removed.
+ */
+export function canRemoveSubscriptionSupervision(
+  extRefOrExtRefs: Element | Element[]
```
JakobVogelsang, line 96: If you want to make this function available for users outside the `scl-lib` you need to either 
1. restrict the input to single `ExtRef` element or 
2. extend the signiture of the output to somethink like 
```ts
{
    canBeRemove: boolean,
    toBeRemoved: GroupedExtRefs
}
```

If you really need this function somewhere, I would be in favor of the second option. The first one will lead to mistakes on the user's side as soon as it is used somewhere in a loop/map.

tLN/canRemoveSubscriptionSupervision.ts
```diff
@@ -0,0 +1,109 @@
...
+): boolean {
+  if (!extRefOrExtRefs) return false;
```
JakobVogelsang, line 98: Is that a scenario that need to be covered? This if-clause can never be reached. I wonder why the coveradge did not show this.

index.ts
```diff
@@ -1,3 +1,5 @@
+import { canRemoveSubscriptionSupervision } from "./tLN/canRemoveSubscriptionSupervision.js";
```
JakobVogelsang, line 1: This is exported twice here.

tExtRef/unsubscribe.spec.ts
```diff
@@ -99,17 +106,25 @@ describe("Function allowing to unsubscribe multiple external references", () =>
...
     expect(actions[1]).to.satisfies(isRemove);
     expect((actions[1] as Remove).node).to.equal(extRefs[1]);
     expect(actions[2]).to.satisfies(isUpdate);
     expect((actions[2] as Update).element).to.equal(extRefs[2]);
-    expect(actions[3]).to.satisfies(isRemove);
-    expect((actions[3] as Remove).node).to.equal(doi);
-    expect(actions[4]).to.satisfies(isRemove);
-    expect((actions[4] as Remove).node).to.equal(ln);
+
+    expect(isRemove(actions[3])).to.be.true;
```
JakobVogelsang, line 117: I would not want to have two signatures to check for edit objects. So either you change both to:

`expect(isRemove(...)).to.be.true` or (my favourite) change all to
`expect(...).to.satisfy(isRemove)`

isRmove above stand for isInsert as well

tLN/canRemoveSubscriptionSupervision.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+const docEd2Goose = findElement(gooseSubscriptionEd2) as XMLDocument;
+const docEd2Sv = findElement(svSubscriptionEd2) as XMLDocument;
+
+// const docEd2Sv = findElement(svSubscriptionEd2) as XMLDocument;
+
+describe("Function to check if supervision references can be removed (canRemoveSubscriptionSupervision)", () => {
+  describe("For GOOSE", () => {
+    describe("Only allows supervision removal if declared with valKind/valImport on DTTs/instances", () => {
+      it("does not return any action if neither declared", () => {
+        const extRef = docEd2Goose.querySelector(
```
JakobVogelsang, line 20: You are testing exclusively with single ExtRefs. If you want to make this function available I would be in favour to have more somplex tests with ExtRef arrays in the test. I personally do not see this function useful outside of `removeSubscriptionSupervision` and would then not export it.

tLN/removeSubscriptionSupervision.ts
```diff
@@ -142,21 +63,48 @@ function groupPerControlBlock(
  * @param extRefs - An array of external reference elements
  * @returns Actions to remove subscription supervision `LGOS` or `LSVS`
  */
-export function removeSubscriptionSupervision(extRefs: Element[]): Remove[] {
-  if (extRefs.length === 0) return [];
+export function removeSubscriptionSupervision(
```
JakobVogelsang, line 66: I am a pain in the ass with this PR. I am really sorry. This API as well does not really feel correct to me. I can imagine that users that want to remove subscription supervision are not intuitively having the ExtRef available. You know, I would rather think of something like this:

`removeSubscriptionSupervision(controlBlock: Element, subscriberIed: Element)`. This is still doing some useful things as the check for `valImport` and `valKind`. You could also think of having an option that allows you to define an array of ExtRefs to be checked against, like so:

```ts
RemoveSubscrptionSupervisionOptions = {
    /** Whether to only empty the Val text content */
    valOnly: boolean;
    /** ExtRef elements triggering this remove default [] means direct remove */
    pendingUnsubscription: Element[];
}
```
option: {valOnly: true, pendingUnsubscriptions: []}


I think that when someone is using this function, this person is knowing what to do. If not, it should be handled by the unsubscribe function.

tLN/removeSubscriptionSupervision.ts
JakobVogelsang: Obsolete, see comment the line above.

JakobVogelsang (changes requested):
@danyill Your two points that you have formulated in the linked issues are very valid and those are covered here in well. I have a bit pain with the function exports. We need to talk through the API I think before you can proceed here.

tLN/canRemoveSubscriptionSupervision.spec.ts
danyill: There are some tests with multiple ExtRefs

tLN/canRemoveSubscriptionSupervision.ts
danyill: I'd like a function like this so I can show in a UI whether it is possible to remove a subscription supervision.

danyill:
Thank you for your review. If you agree, let's close this PR and agree the appropriate API and then perhaps code which may be of use from here can be transferred into a new, clean PR. 

I had, perhaps mistakenly thought since I was adding to tests to an already existing piece of code we had already agreed on this but I think that was not an appropriate assumption.

JakobVogelsang:
Let's close. Sorry for the additional effort, I am causing you.

Issue #20, opened 2023-06-02
(tLN) (tControl) Instantiate Supervision


danyill:
We currently have:

```ts
/**
 * Returns an array with a single Insert Edit to create a new
 * supervision element for the given GOOSE/SMV message and subscriber IED.
 *
 * @param controlBlock The GOOSE or SMV message element
 * @param subscriberIED The subscriber IED
 * @returns an empty array if instantiation is not possible or an array with a single Create action
 */
export function instantiateSubscriptionSupervision(
  controlBlock: Element | undefined,
  subscriberIED: Element | undefined
): Insert[] {
  const supervisionType =
    controlBlock?.tagName === 'GSEControl' ? 'LGOS' : 'LSVS';
  if (
    !controlBlock ||
    !subscriberIED ||
    !isSupervisionAllowed(controlBlock, subscriberIED, supervisionType)
  )
    return [];
  const availableLN = findOrCreateAvailableLNInst(
    controlBlock,
    subscriberIED,
    supervisionType
  );
  if (
    !availableLN ||
    !isSupervisionModificationAllowed(subscriberIED, supervisionType)
  )
    return [];

  const edits: Insert[] = [];
  // If creating new LN element
  if (!availableLN.parentElement) {
    const parent = subscriberIED.querySelector(
      `LN[lnClass="${supervisionType}"]`
    )?.parentElement;
    if (parent) {
      // use Insert edit for supervision LN
      edits.push({
        parent,
        node: availableLN,
        reference:
          parent!.querySelector(`LN[lnClass="${supervisionType}"]:last-child`)
            ?.nextElementSibling ?? null,
      });
    }
  }

  // Insert child elements
  const supervisionName = supervisionType === 'LGOS' ? 'GoCBRef' : 'SvCBRef';

  let doiElement = availableLN.querySelector(`DOI[name="${supervisionName}"]`);
  if (!doiElement) {
    doiElement = subscriberIED.ownerDocument.createElementNS(
      SCL_NAMESPACE,
      'DOI'
    );
    doiElement.setAttribute('name', supervisionName);
    edits.push({
      parent: availableLN!,
      reference: null,
      node: doiElement,
    });
  }

  let daiElement = doiElement.querySelector(`DAI[name="setSrcRef"]`);
  if (!daiElement) {
    daiElement = subscriberIED.ownerDocument.createElementNS(
      SCL_NAMESPACE,
      'DAI'
    );
[...]

PR #40, opened 2023-08-01
feat: add check for subscription  supervision instantiation

JakobVogelsang:
This function is a first step towards an automated subscription supervision after subscription itself. @danyill convinced me that it might be beneficial to have the check exported independent of the instantiation of the supervision.

There is no issue for this as it dropped-off the ongoing work on #20

### DIFF
--- a/tLN/canInstantiateSubscriptionSupervision.ts
+++ b/tLN/canInstantiateSubscriptionSupervision.ts
@@ -1,0 +1,211 @@
+import { controlBlockObjRef } from "../tControl/controlBlockObjRef.js";
+
+export type Supervision = {
+  /** Pointer to the supervision location. This can be either a subscriber IED
+   * or the logical node (`LGOS`/`LSVS`) itself */
+  subscriberIedOrLn: Element;
+  /** The control block to be supervised */
+  sourceControlBlock: Element;
+};
+
+export type SupervisionOptions = {
+  /** Whether to check for `valKind`/`valImport`. Defaulting to true */
+  checkEditableSrcRef: boolean;
+  /**
+   * Whether the given control block is already supervised in the subscriber
+   * IED. Defaulting to true.
+   */
+  checkDuplicateSupervisions: boolean;
+  /**
+   * Whether a new supervision would exceed the limits set in the `Services`
+   * element or whether the subscriber LN already hosts a valid supervision.
+   * Defaulting to true.
+   */
+  checkMaxSupervisionLimits: boolean;
+};
+
+function supervisionLnClass(supervision: Supervision): "LGOS" | "LSVS" {
+  const serviceType = supervision.sourceControlBlock.tagName;
+  return serviceType === "GSEControl" ? "LGOS" : "LSVS";
+}
+
+function type(supervision: Supervision): "GoCBRef" | "SvCBRef" {
+  const serviceType = supervision.sourceControlBlock.tagName;
+  return serviceType === "GSEControl" ? "GoCBRef" : "SvCBRef";
+}
+
+/** @returns Whether a supervision LN holds a valid control block object ref */
+function holdsValidObjRef(ln: Element, type: "GoCBRef" | "SvCBRef"): boolean {
+  const objRef = ln.querySelector(
+    `:scope > DOI[name="${type}"] > DAI[name="setSrcRef"] > Val`
+  )?.textContent;
+  if (!objRef) return false;
+
+  // IEDnameLDinst/prefixLnClassLnInst.CbName
+  const indexSlash = objRef.indexOf("/");
+  const indexDot = objRef.indexOf(".");
+  return (
+    indexSlash !== 0 &&
+    indexSlash < indexDot + 1 &&
+    indexDot - 1 < objRef.length
+  );
+}
+
+/** @returns Whether `Services` element requirement is met */
+function exceedSupervisionLimits(supervision: Supervision): boolean {
+  const subscriberIed =
+    supervision.subscriberIedOrLn.tagName === "IED"
+      ? supervision.subscriberIedOrLn
+      : supervision.subscriberIedOrLn.closest("IED")!;
+
+  const lnClass = supervisionLnClass(supervision);
+  const max = subscriberIed
+    ?.querySelector("Services > SupSubscription")
+    ?.getAttribute(`${lnClass === "LGOS" ? "maxGo" : "maxSv"}`);
+  if (!max || isNaN(parseInt(max, 10))) return false;
+
+  const existingSupervisionLogicalNode = Array.from(
+    subscriberIed.querySelectorAll(`LN[lnClass="${lnClass}"]`)
+  );
+
+  const availableSupervisorSpots = existingSupervisionLogicalNode.filter(
+    (ln) => !holdsValidObjRef(ln, type(supervision))
+  );
+
+  return (
+    existingSupervisionLogicalNode.length <= parseInt(max, 10) &&
+    availableSupervisorSpots.length > 0
+  );
+}
+
+/** @returns Whether `DA`|`DAI` with name `setSrcRef` is editable */
+function isSrcRefEditable(supervision: Supervision): boolean {
+  const lnClass = supervisionLnClass(supervision);
+  const ln =
+    supervision.subscriberIedOrLn.tagName === "LN"
+      ? supervision.subscriberIedOrLn
+      : supervision.subscriberIedOrLn.querySelector(
+          `LN[lnClass="${lnClass}"]`
+        )!;
+
+  const doiName = type(supervision);
+  if (
+    ln.querySelector(
+      `:scope > DOI[name="${doiName}"] > 
+        DAI[name="setSrcRef"][valImport="true"][valKind="RO"],
+       :scope > DOI[name="${doiName}"] > 
+        DAI[name="setSrcRef"][valImport="true"][valKind="Conf"]`
+    )
+  )
+    return true;
+
+  const rootNode = ln.ownerDocument;
+
+  const lnType = ln.getAttribute("lnType");
+
+  const goOrSvCBRef = rootNode.querySelector(
+    `DataTypeTemplates > 
+            LNodeType[id="${lnType}"][lnClass="${lnClass}"] > DO[name="${type(
+      supervision
+    )}"]`
+  );
+
+  const cbRefId = goOrSvCBRef?.getAttribute("type");
+  const setSrcRef = rootNode.querySelector(
+    `DataTypeTemplates > DOType[id="${cbRefId}"] > DA[name="setSrcRef"]`
+  );
+
+  return (
+    (setSrcRef?.getAttribute("valKind") === "Conf" ||
+      setSrcRef?.getAttribute("valKind") === "RO") &&
+    setSrcRef.getAttribute("valImport") === "true"
+  );
+}
+
+/**
+ * A pre requirement for a subscription supervision is an already existing
+ * LN of the class `LGOS` or `LSVS` in the subscriber IED.
+ * @returns Whether there is a logical node element `LN` with the class `LGOS`, `LSVS`
+ */
+function existFirstSupervisionOfType(supervision: Supervision): boolean {
+  const lnClass =
+    supervision.sourceControlBlock.tagName === "GSEControl" ? "LGOS" : "LSVS";
+
+  const firstSupervisionOfType = supervision.subscriberIedOrLn.querySelector(
+    `:scope > AccessPoint > Server > LDevice > LN[lnClass="${lnClass}"]`
+  );
+
+  return firstSupervisionOfType ? true : false;
+}
+
+/** @returns Whether [[`controlBlock`]] is supervised in [[`subscriberIed`]] */
+function isControlBlockSupervised(supervision: Supervision): boolean {
+  const subscriberIed =
+    supervision.subscriberIedOrLn.tagName === "IED"
+      ? supervision.subscriberIedOrLn
+      : supervision.subscriberIedOrLn.closest("IED")!;
+
+  const lnClass =
+    supervision.sourceControlBlock.tagName === "GSEControl" ? "LGOS" : "LSVS";
+  const refType =
+    supervision.sourceControlBlock.tagName === "GSEControl"
+      ? "GoCBRef"
+      : "SvCBRef";
+
+  return Array.from(
+    subscriberIed?.querySelectorAll(
+      `:scope > AccessPoint > Server > LDevice > LN[lnClass="${lnClass}"] 
+      > DOI[name="${refType}"] > DAI[name="setSrcRef"] > Val`
+    )
+  ).some(
+    (val) =>
+      val.textContent === controlBlockObjRef(supervision.sourceControlBlock)
+  );
+}
+
+/** Whether subscription supervision instantiation can be performed.
+ * ```md
+ * - check if `valImport` and `valKind` allow to change subscriber logical node
+ * - check whether the control block is already supervised in the IED
+ * - check whether there is an available location for the control block
+ *   reference to be stored in the supervision source reference.
+ * - check whether `Service` element requirements are met
+ * - check whether the logical node has missing or empty `Val` content
+ *   (iedOrLn is LN)
+ * ```
+ * @returns Whether subscription supervision can be done */
+export function canInstantiateSubscriptionSupervision(
+  supervision: Supervision,
+  options: SupervisionOptions = {
+    checkEditableSrcRef: true,
+    checkDuplicateSupervisions: true,
+    checkMaxSupervisionLimits: true,
+  }
+): boolean {
+  if (
+    options.checkDuplicateSupervisions &&
+    isControlBlockSupervised(supervision)
+  )
+    return false;
+
+  if (supervision.subscriberIedOrLn.tagName === "LN") {
+    const type =
+      supervision.sourceControlBlock.tagName === "GSEControl"
+        ? "GoCBRef"
+        : "SvCBRef";
+    if (holdsValidObjRef(supervision.subscriberIedOrLn, type)) return false;
+  } else {
+    if (!existFirstSupervisionOfType(supervision)) return false;
+  }
+
+  if (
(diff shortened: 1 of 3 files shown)


danyill:
Oops, I've missed this - I'll look at it soon!

danyill:
> Oops, I've missed this - I'll look at it soon!

It really wasn't soon, I am sorry. However I am looking at it right now :wink:

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,134 @@
...
+ * A pre requirement for a subscription supervision is an already existing
+ * LN of the class `LGOS` or `LSVS` in the subscriber IED.
+ * @returns Whether there is a logical node element `LN` with the class `LGOS`, `LSVS`
+ */
+function existFormLogicalNode(supervision: Supervision): boolean {
+  const lnClass =
+    supervision.sourceControlBlock.tagName === "GSEControl" ? "LGOS" : "LSVS";
+
+  const formLn = supervision.subscriberIedOrLn.querySelector(
+    `LDevice > LN[lnClass="${lnClass}"]`
```
danyill, line 87: How about scoping more tightly...

```suggestion
    `:scope > AccessPoint > Server > LDevice > LN[lnClass="${lnClass}"]`
```

One day performance will matter more....

danyill (changes requested):
Thank you for working on this functionality :heart_eyes: which I'm looking forward to incorporating into my supervision editor. You are a :star2: 

For now I've just done an initial view by reading the code and have made a few comments, which I'm happy to discuss.

I note :notebook: that for this call signature to be correct, modifying a supervision can only mean either modifying the existing LN (e.g. removing or adding textContent value) or adding a new LN.

It cannot mean removing an LN completely (as we have done historically) because then we'd need to at least be considering refusing to remove the first instantiated supervision in our checks one. But I think we have already agreed on this, just mentioning because I can.

I've also had a look at how this function would integrate into my oscd-supervision plugin where it would be a drop-in replacement (I think) for `isSupervisionModificationAllowed`  and it would be convenient if a control block was not specified that the other checks were still done (i.e. to answer the question "Is this IED the type of thing which could have supervisions instantiated?"). 

Would it be a useful change to make `sourceControlBlock` optional to meet this use case?

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,134 @@
...
+ * - check whether `Service` element requirements are met
+ * ```
+ * @returns Whether subscription supervision can be done */
+export function canInstantiateSubscriptionSupervision(
+  supervision: Supervision
+): boolean {
+  if (supervision.subscriberIedOrLn.tagName === "LN")
+    return (
+      !isControlBlockSupervised(supervision) &&
+      !supervision.subscriberIedOrLn.querySelector("Val")?.textContent &&
```
danyill, line 123: I don't think we can just look for any `Val` within the LN like this. It needs to be scoped more tightly.
But even if there is no `Val` element, wouldn't we keep our existing behaviour and add a DOI/DAI/Val as required?

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,134 @@
...
+ * ```md
+ * The
+ * - check if `valImport` and `valKind` allow to change subscriber logical node
+ * - check whether the control block is already supervised in the IED
+ * - check whether there are as a given location for subscriber logical node (iedOrLn is IED)
+ * - check whether the logical node has missing or empty val content (iedOrLn is LN)
+ * - check whether `Service` element requirements are met
+ * ```
+ * @returns Whether subscription supervision can be done */
+export function canInstantiateSubscriptionSupervision(
```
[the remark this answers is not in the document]
JakobVogelsang: Thank you :)

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,134 @@
...
+    setSrcRef.getAttribute("valImport") === "true"
+  );
+}
+
+/**
+ * A pre requirement for a subscription supervision is an already existing
+ * LN of the class `LGOS` or `LSVS` in the subscriber IED.
+ * @returns Whether there is a logical node element `LN` with the class `LGOS`, `LSVS`
+ */
+function existFormLogicalNode(supervision: Supervision): boolean {
```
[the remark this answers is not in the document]
JakobVogelsang: I would prefer `existFirstSupervisionOfType`. This comes closer to what we are looking for here.

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/** @returns Whether [[`controlBlock`]] is supervised in [[`subscriberIed`]] */
+function isControlBlockSupervised(supervision: Supervision): boolean {
+  const subscriberIed =
+    supervision.subscriberIedOrLn.tagName === "IED"
+      ? supervision.subscriberIedOrLn
+      : supervision.subscriberIedOrLn.closest("IED")!;
+
+  return Array.from(subscriberIed?.querySelectorAll(`:scope LN Val`)).some(
```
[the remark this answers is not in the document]
JakobVogelsang: Very true thank you!

tLN/canInstantiateSubscriptionSupervision.ts
JakobVogelsang: true

tLN/canInstantiateSubscriptionSupervision.ts
JakobVogelsang: Hmmm this and the other two issue you found are pointing in the same direction as that there are use cases where a user want to overwrite and existing supervision and still want to know whether that can be done in the given SCL. 

I would propose this to make the function more modular:
type CheckSupervisionOption = {
   /** Whether attributes `valKind` and `valImport` allow changing subscription: Defaults to true */
   checkSrcRefEdiable: boolean;
   /** Whether the control block within supervision is already subscribed: Defaults to true */
   checkCtrlBlockSubscribed: boolean;
   /** Whether there are "empty" spots to subscribe to: Defaults to true */
   checkMaxSupervisionLimits: boolean;


That would allow you to not check for the last one in scenarios you want to overwrite a supervision instead of instantiate a new one.

@danyill what are your thoughts on this?

danyill: Sounds good to me, thank you for your consideration :+1:

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,134 @@
...
+ * ```
+ * @returns Whether subscription supervision can be done */
+export function canInstantiateSubscriptionSupervision(
+  supervision: Supervision
+): boolean {
+  if (supervision.subscriberIedOrLn.tagName === "LN")
+    return (
+      !isControlBlockSupervised(supervision) &&
+      !supervision.subscriberIedOrLn.querySelector("Val")?.textContent &&
+      isSupervisionAllowed(supervision) &&
```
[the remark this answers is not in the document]
JakobVogelsang: The way `isSpuervisionAllowed` works is to count the number of all already in use supervisions. It does not take into account any logical nodes LGOS / LSVS with missing Val with DAI setSrcRef ... . What I will do as well is to make sure it does also not take into account invalid string in `Val`. There must be a `/` followed by a `.`. 

Do you see any used cases where you want to overwrite those? I would intuitively say no! In the plugin, it would make sense to first remove the supervision before adding new control blocks to it. WDYT?

JakobVogelsang:
WDYT now? I have added some test on the edge cases you might looking for.

tLN/canInstantiateSubscriptionSupervision.ts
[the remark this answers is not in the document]
danyill: I'm not so sure about the invalid string but I have no fundamental objections, I guess it's a way of saying "is it a valid object reference?".

I think it's a reasonable restriction to require removal before adding.

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,191 @@
...
+  const type = serviceType === "GSEControl" ? "GoCBRef" : "SvCBRef";
+
+  const max = subscriberIed
+    ?.querySelector("Services > SupSubscription")
+    ?.getAttribute(`${lnClass === "LGOS" ? "maxGo" : "maxSv"}`);
+  if (!max || isNaN(parseInt(max, 10))) return false;
+
+  const existingSupervision = Array.from(
+    subscriberIed.querySelectorAll(`LN[lnClass="${lnClass}"]`)
+  ).filter((ln) => holdsValidObjRef(ln, type)).length;
```
danyill, line 61: I'm a bit confused as to why we shouldn't just take the length of the first query here. Why should we check to see if we have a valid object reference?

In other words, why isn't the following sufficient?

```ts
const existingSupervision = Array.from(
    subscriberIed.querySelectorAll(`LN[lnClass="${lnClass}"]`)
  )
```

I see that what we're doing here is similar to how we're doing supervisions previously in the subscriber plugins but I'm not entirely convinced it's correct.

We are checking to see if we are able to instantiate a new LGOS or LSVS LN and for that purpose we only care (I think?!) about the total number of LGOS and LSVS compared to the totals in maxGo and maxSv.

I feel like I've not understood something here.

tLN/canInstantiateSubscriptionSupervision.ts
```diff
@@ -0,0 +1,191 @@
...
+      ? supervision.subscriberIedOrLn
+      : supervision.subscriberIedOrLn.querySelector(
+          `LN[lnClass="${lnClass}"]`
+        )!;
+
+  if (
+    ln.querySelector(
+      ':scope DAI[name="setSrcRef"][valImport="true"][valKind="RO"],' +
+        ' :scope DAI[name="setSrcRef"][valImport="true"][valKind="Conf"]'
+    )
```
danyill, line 82: This is not tight enough. We have to specify the DOI as well.

See below for example, there are multiple `setSrcRef` in the following LGOS LN, and while it'll probably select the right one, I think I had some problems with this previously.

```xml
<LN lnType="LGOS_4" lnClass="LGOS" inst="1" prefix="">
	<Private type="GE_Digital_Energy_UR LN_UUID">Master_LGOS1</Private>
	<DOI name="NamPlt">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Ctl</Val>
		</DAI>
	</DOI>
	<DOI name="Beh">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>This logical node's behaviour</Val>
		</DAI>
		<DAI name="stVal" valKind="RO"/>
	</DOI>
	<DOI name="St">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>RxGOOSE1 On operand</Val>
		</DAI>
		<DAI name="stVal" valKind="RO" sAddr="0:1441793,0"/>
	</DOI>
	<DOI name="Alm">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>RxGOOSE1 Off operand</Val>
		</DAI>
		<DAI name="stVal" valKind="RO" sAddr="0:1507329,0"/>
	</DOI>
	<DOI name="SimSt">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Status showing that Sim messages are being received and accepted for RxGOOSE1</Val>
		</DAI>
		<DAI name="stVal" valKind="RO" sAddr="2:288538,0"/>
	</DOI>
	<DOI name="LastStNum">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Actual value RxGOOSE1 StNum</Val>
		</DAI>
		<DAI name="stVal" valKind="RO" sAddr="1:316928,0"/>
		<SDI name="units">
			<DAI name="SIUnit" valKind="RO">
				<Val/>
			</DAI>
			<DAI name="multiplier" valKind="RO">
				<Val/>
			</DAI>
		</SDI>
	</DOI>
	<DOI name="LastSqNum">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Actual value RxGOOSE1 SqNum</Val>
		</DAI>
		<DAI name="stVal" valKind="RO" sAddr="1:316930,0"/>
		<SDI name="units">
			<DAI name="SIUnit" valKind="RO">
				<Val/>
			</DAI>
			<DAI name="multiplier" valKind="RO">
				<Val/>
			</DAI>
		</SDI>
	</DOI>
	<DOI name="RxConfRevNum">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Status of the confRev field in last accepted RxGOOSE1 message</Val>
		</DAI>
		<DAI name="stVal" valKind="RO" sAddr="1:316932,0"/>
		<SDI name="units">
			<DAI name="SIUnit" valKind="RO">
				<Val/>
			</DAI>
			<DAI name="multiplier" valKind="RO">
				<Val/>
			</DAI>
		</SDI>
	</DOI>
	<DOI name="ConfRevNum">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Expected ConfRev number</Val>
		</DAI>
		<DAI name="stVal" valKind="RO" valImport="true">
			<Val>1</Val>
		</DAI>
		<SDI name="units">
			<DAI name="SIUnit" valKind="RO">
				<Val/>
			</DAI>
			<DAI name="multiplier" valKind="RO">
				<Val/>
			</DAI>
		</SDI>
	</DOI>
	<DOI name="GoCBRef">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Setting RxGOOSE1 GoCBRef</Val>
		</DAI>
		<DAI name="setSrcRef" valKind="RO" valImport="true">
			<Val></Val>
		</DAI>
	</DOI>
	<DOI name="MAC">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Destination MAC address of messages subscribed to by RxGOOSE1</Val>
		</DAI>
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val>00-00-00-00-00-00</Val>
		</DAI>
	</DOI>
	<DOI name="RgRxMode">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>R-RxGOOSE1 RECEPTION MODE setting</Val>
		</DAI>
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val>SSM</Val>
		</DAI>
	</DOI>
	<DOI name="RgSrcIP">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Source IP address of subscribed R-GOOSE message</Val>
		</DAI>
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val>127.0.0.1</Val>
		</DAI>
	</DOI>
	<DOI name="RgDstIP">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Destination IP address of subscribed R-GOOSE message</Val>
		</DAI>
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val>224.0.0.0</Val>
		</DAI>
	</DOI>
	<DOI name="SecEna">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>R-RxGOOSE1 SECURITY setting</Val>
		</DAI>
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val>None</Val>
		</DAI>
	</DOI>
	<DOI name="APPID">
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val>0</Val>
		</DAI>
		<SDI name="units">
			<DAI name="SIUnit" valKind="RO">
				<Val/>
			</DAI>
			<DAI name="multiplier" valKind="RO">
				<Val/>
			</DAI>
		</SDI>
		<DAI name="minVal" valKind="RO">
			<Val>0</Val>
		</DAI>
		<DAI name="maxVal" valKind="RO">
			<Val>65535</Val>
		</DAI>
		<DAI name="stepSize" valKind="RO">
			<Val>1</Val>
		</DAI>
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Setting RxGOOSE1 ETYPE APPID</Val>
		</DAI>
	</DOI>
	<DOI name="GoDatSetRef">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Setting RxGOOSE1 DatSet</Val>
		</DAI>
		<DAI name="setSrcRef" valKind="RO" valImport="true">
			<Val></Val>
		</DAI>
	</DOI>
	<DOI name="GoID">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>Setting RxGOOSE1 goID</Val>
		</DAI>
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val></Val>
		</DAI>
	</DOI>
	<DOI name="RxMode">
		<DAI name="d" valKind="RO" valImport="true">
			<Val>RxGOOSE1 MODE setting</Val>
		</DAI>
		<DAI name="setVal" valKind="RO" valImport="true">
			<Val>Main CPU GOOSE</Val>
		</DAI>
	</DOI>
</LN>
```

danyill (changes requested):
It's looking good, thank you for the discussion and for looking through my comments.

I think there is one specifier which is not tight enough and I have one question.

tLN/canInstantiateSubscriptionSupervision.ts
JakobVogelsang: Thank you @danyill for not letting go of this. This was a misunderstanding on my side. For me, `maxGo` and `maxSv` clearly were the maximum subscription supervisions. With that definition an LGOS with even an empty Val would not count and this is how I did implement this function. The standard however is very clear on that and speaks about `maximum number of subscription supervision LNs`. 

Still it is for us to decide waht the function is checking for. I think this might be a naming issue in its core. When `instantiating` means to create a new LGOS/LSVS of some sort with or without valid supervision then you are right we need to just check for the number of LGOS/LSVS. When instantiating means add a supervision by taking one available LGOS/LSVS or creating a new one, we still need to check for a possible subscription supervision. Is it? WDYT?

tLN/canInstantiateSubscriptionSupervision.ts
JakobVogelsang: Very true thank you!!!

danyill (approved):
LGTM, thank you for tenaciously pursuing this every so slightly tricky set of conditions! :100:

PR #56, opened 2023-10-04
feat(tControl): function for looking up communications address

### LINKED ISSUE #41: Lookup GSE or SMV from control block
I'd be keen to add a lookup of a GSE or SMV address element from a control block if this is agreed to be appropriate for  this library.

The main subtlety is supporting `ServerAt` elements which is easy to forget.

I'm currently using:

```ts
/** @returns the cartesian product of `arrays` */
export function crossProduct<T>(...arrays: T[][]): T[][] {
  return arrays.reduce<T[][]>(
    (a, b) => <T[][]>a.flatMap(d => b.map(e => [d, e].flat())),
    [[]]
  );
}

function getCommAddress(ctrlBlock: Element): Element {
  const doc = ctrlBlock.ownerDocument;

  const ctrlLdInst = ctrlBlock.closest('LDevice')!.getAttribute('inst');
  const addressTag = ctrlBlock.tagName === 'GSEControl' ? 'GSE' : 'SMV';
  const ied = ctrlBlock.closest('IED')!;
  const iedName = ied.getAttribute('name');
  const apName = ctrlBlock.closest('AccessPoint')?.getAttribute('name');

  const cbName = ctrlBlock.getAttribute('name');

  let apNames = [];
  const serverAts = ied.querySelectorAll(
    `AccessPoint > ServerAt[apName="${apName}"`
  );
  if (serverAts) {
    const serverAtNames = Array.from(serverAts).map(ap =>
      ap.closest('AccessPoint')!.getAttribute('name')
    );
    apNames = [apName, ...serverAtNames];
  } else {
    apNames = [apName];
  }

  const connectedAps = `Communication > SubNetwork > ConnectedAP[iedName="${iedName}"]`;
  const connectedApNames = apNames.map(ap => `[apName="${ap}"]`);
  const addressElement = `${addressTag}[ldInst="${ctrlLdInst}"][cbName="${cbName}"]`;

  return doc.querySelector(
    crossProduct([connectedAps], connectedApNames, ['>'], [addressElement])
      .map(strings => strings.join(''))
      .join(',')
  )!;
}
```

I'd be happy to add some tests and contribute this at some point.

danyill:
Closes #41

I would be most grateful for a review and happy to make changes.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -9,6 +9,8 @@ export { removeProcessElement } from "./tSubstation/removeProcessElement.js";
 
 export { InsertIedOptions, insertIed } from "./tIED/insertIED.js";
 
+export { controlBlockGseOrSmv } from "./tControl/controlBlockGseOrSmv.js";
+
 export { CreateReportControlOptions } from "./tReportControl/createReportControl.js";
 export { updateReportControl } from "./tReportControl/updateReportControl.js";
 export { createReportControl } from "./tReportControl/createReportControl.js";
--- a/foundation/utils.ts
+++ b/foundation/utils.ts
@@ -45,3 +45,11 @@ export function createElement(
 
   return element;
 }
+
+/** @returns the cartesian product of `arrays` */
+export function crossProduct<T>(...arrays: T[][]): T[][] {
+  return arrays.reduce<T[][]>(
+    (a, b) => <T[][]>a.flatMap((d) => b.map((e) => [d, e].flat())),
+    [[]]
+  );
+}
--- a/tControl/controlBlockGseOrSmv.spec.ts
+++ b/tControl/controlBlockGseOrSmv.spec.ts
@@ -1,0 +1,110 @@
+import { expect } from "chai";
+
+import { controlBlockGseOrSmv } from "./controlBlockGseOrSmv.js";
+import { controlBlockGseOrSmvDoc } from "./controlBlockGseOrSmv.testfiles.js";
+import { findElement } from "../foundation/helpers.test.js";
+
+const testDoc = findElement(controlBlockGseOrSmvDoc) as XMLDocument;
+
+describe("Function to return a communication address for a control block", () => {
+  it("returns a GSE for a GSEControl with the correct AP", () => {
+    const gseControl = testDoc.querySelector(
+      'IED[name="GOOSE_Publisher"] GSEControl[name="GOOSE2"]'
+    )!;
+    const gseComm = testDoc.querySelector(
+      'ConnectedAP[iedName="GOOSE_Publisher"] GSE[cbName="GOOSE2"]'
+    )!;
+
+    expect(gseControl).to.not.be.null;
+    expect(gseComm).to.not.be.null;
+    expect(controlBlockGseOrSmv(gseControl)).to.equal(gseComm);
+  });
+
+  it("returns a GSE for a GSEControl with the correct AP with no ServerAts defined", () => {
+    const gseControl = testDoc.querySelector(
+      'IED[name="GOOSE_Publisher2"] GSEControl[name="GOOSE2"]'
+    )!;
+    const gseComm = testDoc.querySelector(
+      'ConnectedAP[iedName="GOOSE_Publisher2"] GSE[cbName="GOOSE2"]'
+    )!;
+
+    expect(gseControl).to.not.be.null;
+    expect(gseComm).to.not.be.null;
+    expect(controlBlockGseOrSmv(gseControl)).to.equal(gseComm);
+  });
+
+  it("returns null if a GSEControl has no GSE", () => {
+    const gseControl = testDoc.querySelector(
+      'IED[name="GOOSE_Publisher"] GSEControl[name="GOOSE3"]'
+    )!;
+
+    expect(gseControl).to.not.be.null;
+    expect(controlBlockGseOrSmv(gseControl)).to.equal(null);
+  });
+
+  it("returns a GSE for a GSEControl referenced with ServerAt", () => {
+    const gseControl = testDoc.querySelector(
+      'IED[name="GOOSE_Publisher"] GSEControl[name="GOOSE1"]'
+    )!;
+    const gseComm = testDoc.querySelector(
+      'ConnectedAP[apName="PP1"][iedName="GOOSE_Publisher"] > GSE[cbName="GOOSE1"]'
+    )!;
+
+    expect(gseControl).to.not.be.null;
+    expect(gseComm).to.not.be.null;
+    expect(controlBlockGseOrSmv(gseControl)).to.equal(gseComm);
+  });
+
+  it("returns null for an orphan GSEControl", () => {
+    const orphanGse = `<GSEControl name="GOOSE3" type="GOOSE" appID="GOOSE3" confRev="1" datSet="GOOSE1sDataSet"/>`;
+    const gse = <Element>findElement(orphanGse, "GSEControl");
+
+    expect(gse).to.not.be.null;
+    expect(controlBlockGseOrSmv(gse)).to.be.null;
+  });
+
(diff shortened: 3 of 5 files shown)


index.ts
```diff
@@ -9,6 +9,8 @@ export { removeProcessElement } from "./tSubstation/removeProcessElement.js";
 
 export { InsertIedOptions, insertIed } from "./tIED/insertIED.js";
 
+export { getCommAddress } from "./tControl/getCommAddress.js";
```
JakobVogelsang, line 12: I am not sure if the name of the function is very descriptive. I would have expected that there is some sort of `controlBlock` included. Something like:
- controlBlockCommunicationParameters
- contrlolBlockCommunicationAddress

JakobVogelsang: or event more prices like 

- controlBlockGseOrSmvAddress

tControl/getCommAddress.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Looks up Communication section GSE or SMV addresses based on the control block
+ * within the IED section (GSEControl or SampledValueControl).
+ * @param ctrlBlock - SCL control block element (GSEControl or SampledValueControl)
+ * @returns SCL GSE or SMV address element or null if not found.
+ */
+export function getCommAddress(ctrlBlock: Element): Element | null {
+  const doc = ctrlBlock.ownerDocument;
+
+  const ctrlLdInst = ctrlBlock.closest('LDevice')!.getAttribute('inst');
```
JakobVogelsang, line 18: I would propose to get an early return for orphan control blocks. In this line an error for such things would be the result.

tControl/getCommAddress.ts
```diff
@@ -0,0 +1,49 @@
...
+  const addressTag = ctrlBlock.tagName === 'GSEControl' ? 'GSE' : 'SMV';
+  const ied = ctrlBlock.closest('IED')!;
+
+  const apName = ctrlBlock.closest('AccessPoint')?.getAttribute('name');
+
+  let apNames = [];
+  const serverAts = ied.querySelectorAll(
+    `AccessPoint > ServerAt[apName="${apName}"`
+  );
+  if (serverAts.length > 0) {
```
JakobVogelsang, line 28: Is that really necessary. I think if you process serverAts with a map it would be more readable.

tControl/getCommAddress.ts
```diff
@@ -0,0 +1,49 @@
...
+    const serverAtNames = Array.from(serverAts).map(ap =>
+      ap.closest('AccessPoint')!.getAttribute('name')
+    );
+    apNames = [apName, ...serverAtNames];
+  } else {
+    apNames = [apName];
+  }
+
+  const iedName = ied.getAttribute('name');
```
JakobVogelsang, line 37: I found that function where all the data that is neccassery collected at the beginning are more readable. So

```ts

// collected all data
iedName = ...
ldInst = ...
ied = ...

...
if(!iedName || !ied || !apNames) return null

...

```

tControl/getCommAddress.ts
```diff
@@ -0,0 +1,49 @@
+/** @returns the cartesian product of `arrays` */
+function crossProduct<T>(...arrays: T[][]): T[][] {
```
JakobVogelsang, line 2: Not sure about that, but I would have that rather in the utils file. It does make it more readable and for later contributors makes it easier to find it.

index.ts
JakobVogelsang: I just see that the return is either `GSE` or `SMV`. So maybe `controlBlocksGseOrSvm`. Although the proposals not close to SCL syntax would work for me as well.

tControl/getCommAddress.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+describe("Function to return a communication address for a control block", () => {
+  it("returns a GSE for a GSEControl with the correct AP", () => {
+    const gseControl = testDoc.querySelector(
+      'IED[name="GOOSE_Publisher"] GSEControl[name="GOOSE2"]'
+    )!;
+    const gseComm = testDoc.querySelector(
+      'ConnectedAP[iedName="GOOSE_Publisher"] GSE[cbName="GOOSE2"]'
+    )!;
+
+    expect(gseControl).to.not.equal(null);
```
JakobVogelsang, line 18: chia offers `expect(...).to.not.be.null`

tControl/getCommAddress.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    expect(gseComm).to.not.equal(null);
+    expect(getCommAddress(gseControl)).to.equal(gseComm);
+  });
+
+  it("returns a GSE for a GSEControl with the correct AP with no ServerAts defined", () => {
```
JakobVogelsang, line 23: Test missing with orphan `GSEControl` or `SampledValueControl`

tControl/getCommAddress.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  it("returns null if a GSEControl has no GSE", () => {
+    const gseControl = testDoc.querySelector(
+      'IED[name="GOOSE_Publisher"] GSEControl[name="GOOSE3"]'
+    )!;
+
+    expect(gseControl).to.not.equal(null);
+    expect(getCommAddress(gseControl)).to.equal(null);
+  });
+
+  it("returns a GSE for a GSEControl referenced with ServerAt", () => {
```
JakobVogelsang, line 45: Love those!

JakobVogelsang (changes requested):
The functionality looks very good very good. Thank you for the good tests!!!
Two things I would like to have changed:
- I think we need more edge cases covered. I would at least expect that orphan control blocks are covered. That is not an unlikely scenario.
- the name of the function is not very descriptive. Maybe a refining session would make sence.

tControl/getCommAddress.ts
danyill: Why would we need to cover orphan control blocks? Shouldn't we make some assumptions about our documents, e.g. they are schema valid?

danyill: I'll add it I guess it makes it more robust for weird cases but I'm not sure when to do this and when not to (i.e. how to avoid making pathological assumptions about the document).

tControl/getCommAddress.spec.ts
danyill: Oh that is nicer, thanks

tControl/getCommAddress.spec.ts
danyill: I will add that although I am unsure in what circumstances orphan control blocks can occur. I guess during some manipulation or creation in our code (as opposed to in a file?). Anyway for a little restructuring it seems like a good idea.

danyill:
Thank you for this review, I have endeavoured to make some improvements and taken your suggestion for the function name. I'm not entirely sure when we use `get` or `find` or when we drop the verb entirely...

index.ts
```diff
@@ -9,6 +9,8 @@ export { removeProcessElement } from "./tSubstation/removeProcessElement.js";
 
 export { InsertIedOptions, insertIed } from "./tIED/insertIED.js";
 
+export { controlBlockGseOrSmv as getCommAddress } from "./tControl/controlBlockGseOrSmv.js";
```
JakobVogelsang, line 12: I think this came automatically with a rename. Right or was it your intention?

JakobVogelsang (changes requested):
LGTM aside one little name conflict. Thank you for the changes and for your patience.

index.ts
danyill: Quite right, fixed. Thanks for checking.

JakobVogelsang: Still there nothing changed here

JakobVogelsang (changes requested):
Still the same issue

JakobVogelsang (approved):
LGTM

PR #58, opened 2023-10-10
feat(tIED): updateIED

### LINKED ISSUE #55: tIED: Rename IED
It would be useful to be able to rename an IED.

Renaming an IED in IEC 61850 is not quite so simple because the IED name is used in a range of references.

**Requirements:**

1. Check for existing name duplication. If so, do nothing.
1. Rename the `name` attribute on the IED. Some IED tools set the `desc` attribute to the IED name as well. We _could_ (but probably should not (?) change it here as well).
1. Iterate through the `Substation` section. Update any `iedName` on any relevant `LNode`.
1. Iterate through the `Communication` section. Update the `iedName` within each `ConnectedAP`
1. Within _other_ IEDs, update:
    1. The `iedName` field of the `KDC` element.
    2. The `iedName` field of all `ExtRef` elements.
    3. The `iedName` field of all `ClientLN` elements.
    4. The text content of `IEDName` elements within `GSEControl` and `SampledValueControl` should be replaced by the IED's name. We could (but probably should not) also consider if the `appId` attribute should have the IED name replaced on it as it is common for the `appId` to be prefixed by the IED name.
    5. The `iedName` attribute of the `Association` element (something I have never seen in the wild).
    6. The supervision logical nodes (LSVS or LGOS) control block reference LN > DOI[name=GoCBRef/SvCBRef] > DAI[name= setsrcRef] > Val and then update the text content of the control block reference which is prefixed with the IED name (however we must be careful about ambiguous references).

Quite optionally and depending on the approach taken, we could look at the way other namespaces are used:

We could replace `sIedName` while we are there for IEC 61860-6-100 on `LNodeSpecNaming` elements with an `sIedName` attribute.

```
<Private type="eIEC61850-6-100">
  <eIEC61850-6-100:LNodeInputs>
  <eIEC61850-6-100:SourceRef desc="CB state" service="GOOSE" source="XAT/220kV/B230/DCB232/CB/XCBR4.Pos.stVal" input="CB state" extRefAddr="" pLN="XCBR" pDO="Pos" pDA="stVal" />
  </eIEC61850-6-100:LNodeInputs>
  <eIEC61850-6-100:LNodeSpecNaming sIedName="XAT_232_P2" sLdInst="CB" sPrefix="" sLnClass="XCBR" sLnInst="4" />
</Private>
```

Currently OpenSCD handles many of these:

https://github.com/openscd/open-scd/blob/a737c4211360d2d93d44d1e89e731ac8e4c873f2/packages/open-scd/src/wizards/foundation/references.ts#L25-L52

But does not handle control block references for LSVS and LGOS, see https://github.com/openscd/open-scd/issues/1224

JakobVogelsang:
Closes #55 

After looking at some files I have decided to not have the update of `appId`, `smvId` and `rptId` included as they can differ a lot. We need to find a way on how to modularize this further.  I can think of having an option function that we allow the user to set or maybe the vendors name.

### DIFF
--- a/tIED/updateIED.ts
+++ b/tIED/updateIED.ts
@@ -1,0 +1,165 @@
+import { Edit, Update } from "../foundation/utils.js";
+
+import { controlBlockObjRef } from "../tControl/controlBlockObjRef.js";
+
+function isPublic(element: Element): boolean {
+  return !element.closest("Private");
+}
+
+const elementsWithIedNameAttribute = [
+  "LNode",
+  "ConnectedAP",
+  "KDC",
+  "ExtRef",
+  "ClientLN",
+  "Association",
+];
+
+function updateIEDNameTextContent(
+  ied: Element,
+  oldIedName: string,
+  newIedName: string
+): Edit[] {
+  return Array.from(ied.ownerDocument.getElementsByTagName("IEDName"))
+    .filter(isPublic)
+    .filter((iedName) => iedName.textContent === oldIedName)
+    .map((iedName) => {
+      const node = Array.from(iedName.childNodes).find(
+        (node) => node.nodeType === Node.TEXT_NODE
+      )!;
+
+      return [
+        { node },
+        { parent: iedName, node: document.createTextNode(newIedName) },
+      ];
+    });
+}
+
+/** Valid is:
+ * 1. there is an control block in the IED with the name change (ied)
+ * 2. this control block is subscribed in otherIED (iedName, srcLDInst and srcCBName match)
+ * 3. there is a LGOS/LSVS > ... > setSrcRef holding this control block object reference
+ */
+function validSubscriptionSupervision(
+  ied: Element,
+  otherIED: Element,
+  oldIedName: string
+): Element[] {
+  // for GSEControl elements
+  const lgosVals = Array.from(ied.querySelectorAll("GSEControl"))
+    .filter((srcCB) => {
+      //filter out all control blocks that are not subscribed in otherIED
+      const srcLDInst = srcCB.closest("LDevice")?.getAttribute("inst");
+      return !!otherIED.querySelector(
+        `:scope > AccessPoint > Server > LDevice 
+        ExtRef[iedName="${oldIedName}"][srcLDInst="${srcLDInst}"][srcCBName="${srcCB.getAttribute(
+          "name"
+        )}"]`
+      );
+    })
+    .map((srcCB) => {
+      const objRef = controlBlockObjRef(srcCB);
+      return Array.from(
+        otherIED.querySelectorAll(
+          `:scope > AccessPoint > Server > LDevice > LN[lnClass="LGOS"] > 
+            DOI[name="GoCBRef"] > DAI[name="setSrcRef"] > Val`
+        )
+      ).find((val) => val.textContent === objRef);
+    })
+    .filter((val) => val) as Element[];
+
+  // for SampledValueControl elements
+  const lsvsVals = Array.from(ied.querySelectorAll("SampledValueControl"))
+    .filter((srcCB) => {
+      //filter out all control blocks that are not subscribed in otherIED
+      const srcLDInst = srcCB.closest("LDevice")?.getAttribute("inst");
+      return !!otherIED.querySelector(
+        `:scope > AccessPoint > Server > LDevice 
+          ExtRef[iedName="${oldIedName}"][srcLDInst="${srcLDInst}"][srcCBName="${srcCB.getAttribute(
+          "name"
+        )}"]`
+      );
+    })
+    .map((srcCB) => {
+      const objRef = controlBlockObjRef(srcCB);
+      return Array.from(
+        otherIED.querySelectorAll(
+          `:scope > AccessPoint > Server > LDevice > LN[lnClass="LSVS"] > 
+          DOI[name="SvCBRef"] > DAI[name="setSrcRef"] > Val`
+        )
+      ).find((val) => val.textContent === objRef);
+    })
+    .filter((val) => val) as Element[];
+
+  return [...lgosVals, ...lsvsVals];
+}
+
+function updateSubscriptionSupervision(
+  ied: Element,
+  oldIedName: string,
+  newIedName: string
+): Edit[] {
+  const vals = Array.from(
+    ied.ownerDocument.querySelectorAll(":root > IED")
+  ).flatMap((otherIED) =>
+    validSubscriptionSupervision(ied, otherIED, oldIedName)
+  );
+
+  return vals.flatMap((val) => {
+    const oldContent = val.textContent!;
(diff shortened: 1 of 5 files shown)


tIED/updateIED.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Function to schema valid update name attribute in IED element.
+ * The function makes sure to also
+ * 1. Update all elements with iedName attribute referenced to the IED.name attribute
+ *    such as LNode, ClientLN, ExtRef, KDC, Association, ConnectedAP
+ * 2. Update all control block object references pointing to the IED
+ * 3. Updates IEDName elements text content
+ * @param update - update edit for the IED attributes
+ * @returns - Set of edits including the input update
+ */
+export function updateIED(update: Update): Edit[] {
```
danyill, line 129: It took me a while to understand this call signature and now I do I think it's a good idea and a sound approach. But for a new user to this library I think it could be initially frustrating.

Could we include an example in the docstring? Perhaps something like:

```md

To use, pass an Update edit on an IED element with the name attribute.

For example, to replace an IED, `ied` of name, `TEMPLATE` with `IED1`:

updateIED({element: ied, attributes: {name: 'IED1'}})

```

tIED/updateIED.ts
```diff
@@ -0,0 +1,144 @@
...
+  otherIED: Element,
+  oldIedName: string
+): Element[] {
+  // for GSEControl elements
+  const lgosVals = Array.from(ied.querySelectorAll("GSEControl"))
+    .filter((srcCB) => {
+      //filter out all control blocks that are not subscribed in otherIED
+      const srcLDInst = srcCB.closest("LDevice")?.getAttribute("inst");
+      return !!otherIED.querySelector(
+        `ExtRef[iedName="${oldIedName}"][srcLDInst="${srcLDInst}"][srcCBName="${srcCB.getAttribute(
```
danyill, line 45: Same concern about fully qualified selectors where possible.

danyill, line 45: I guess we don't need `srcLNClass`, `srcLNInst` and `srcPrefix` as there is only one LN0 per LD and for GSEControl elements these are well defined. Just noting this in passing.

tIED/updateIED.ts
```diff
@@ -0,0 +1,144 @@
...
+        )}"]`
+      );
+    })
+    .map((srcCB) => {
+      const objRef = controlBlockObjRef(srcCB);
+      return Array.from(
+        otherIED.querySelectorAll(
+          `LDevice > LN[lnClass="LGOS"] > DOI[name="GoCBRef"] > DAI[name="setSrcRef"] > Val`
+        )
+      ).find((val) => val.textContent === objRef);
```
danyill, line 56: We assume only one supervision is possible per control block in the receiving IED. I think this is correct. I am not sure I have read in the standard that it must be this way but I think it is a reasonable assumption.

tIED/updateIED.ts
```diff
@@ -0,0 +1,144 @@
...
+    })
+    .filter((val) => val) as Element[];
```
danyill, line 58: What is this last filter statement doing?

Why couldn't we end line 57 with:

```
}) as Element [];
```

tIED/updateIED.ts
danyill: I'm not entirely convinced it shouldn't still be called `updateIEDName` :wink:

tIED/updateIED.ts
```diff
@@ -0,0 +1,144 @@
...
+): Edit[] {
+  const vals = Array.from(ied.ownerDocument.querySelectorAll("IED"))
+    .filter((otherIED) => otherIED !== ied)
+    .flatMap((otherIED) =>
+      validSubscriptionSupervision(ied, otherIED, oldIedName)
+    );
+
+  return vals.flatMap((val) => {
+    const oldContent = val.textContent!;
+    const newContent = oldContent.replace(oldIedName, newIedName);
```
danyill, line 97: I guess using `replace` is fine.

Perhaps it is paranoia given the prior checks but I guess I would have simply stripped the length of `oldIedName` off `oldContent`

```ts
`${newIedName}${oldContent.slice(oldIedName.length)}`
```

danyill (approved):
Thank you for the opportunity to review this and for the speed with which you developed this very useful functionality :rocket: 

I think it looks really good and the test coverage, naming and functional breakdown looks tidy.

It may be worth considering:

* perhaps some selectors aren't being scoped as tightly as they could be
* the identification of the textContent could be incorrect under some rare circumstances possibly.

But these are minor niggles, it is a fine piece of work! :1st_place_medal:

tIED/updateIED.ts
```diff
@@ -0,0 +1,144 @@
...
+function updateIEDNameTextContent(
+  ied: Element,
+  oldIedName: string,
+  newIedName: string
+): Edit[] {
+  return Array.from(ied.ownerDocument.querySelectorAll("IEDName"))
+    .filter((iedName) => iedName.textContent === oldIedName)
+    .map((iedName) => {
+      return [
+        { node: iedName.childNodes[0] },
```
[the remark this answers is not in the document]
JakobVogelsang: Very true. I will add to the tests to cover that.

tIED/updateIED.ts
```diff
@@ -0,0 +1,144 @@
...
+): Update[] {
+  const selector = elementsWithIedNameAttribute
+    .map((iedNameElement) => `${iedNameElement}[iedName="${oldIedName}"]`)
+    .join(",");
+
+  return Array.from(ied.ownerDocument.querySelectorAll(selector)).map(
+    (element) => {
+      return { element, attributes: { iedName: newIedName } };
+    }
+  );
```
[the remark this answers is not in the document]
JakobVogelsang: I would say for every query that is not specific. So looking for `IED` in the whole document is a known issue with ABB SCL files as they use this tag name also within a private section.

tIED/updateIED.ts
JakobVogelsang: Yes. I do not want to check the implementation of ExtRef here but do a best guess for the rename

tIED/updateIED.ts
JakobVogelsang: Then we would export (Element | undefined)[]. So just to make the linter happy

tIED/updateIED.ts
JakobVogelsang: It will do the same thing. replace is useing the first find and as this is always at the beginning those two sollutions do the same thing. replace reads a bit better if you ask me.

tIED/updateIED.ts
JakobVogelsang: I did add a lot more description about the function. In general, I think I need to invest some time to explain the idea behind those calls. Once understood, I think it becomes very easy to navigate the library. Thank you for this comment.

PR #66, opened 2023-11-08
feat(tExtRef): add options to fdcaToExtRefCheck

### LINKED ISSUE #65: Provide override of preconfigured checks and only check basic type
We discovered that for some devices there is a need to connect different CDCs, e.g. ACT to SPS.

Currently scl-lib only allows this with the `force` option but this takes away many checks.

@JakobVogelsang and I discussed this morning and we agreed that a "Check only basic type match" option was required.

This means that we wouldn't do checks on pLN or pDO (for CDC), probably we would still check pServT.

My feeling is that the name "Check only basic type match" is still the wrong way to phrase this, and that we should call the option (and the UI) something more along the lines of "Skip Common Data Class Match" or something like this which is more targeted to the actual concern.

More information: https://github.com/danyill/oscd-subscriber-later-binding/issues/10

JakobVogelsang:
Closes #65

### DIFF
--- a/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
+++ b/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
@@ -14,96 +14,148 @@ const extRefXSWIBehQ = `<ExtRef pDO="Beh" pDA="q" pServT="GOOSE" pLN="CSWI"/>`;
 const invalidExtRef = `<ExtRef pDO="someInvalidDO" pDA="Dbpos" pServT="GOOSE" pLN="XSWI"/>`;
 const invalidExtRef1 = `<ExtRef pDA="Dbpos" pServT="GOOSE" pLN="XSWI"/>`;
 const extRefCMVFLOAT32 = `<ExtRef pDO="A.phsA" pDA="cVal.mag.f" pServT="Report" pLN="MMXU"/>`;
+const extRefBlkOpnstVal = `<ExtRef pDO="BlkOpn" pDA="stVal" pLN="CILO" />`;
 
 describe("Function compare FCDA basic types to ExtRef restrictions", () => {
   it("is restrictive with invalid FCDA input", () => {
-    const fcda = findElement(orphanFCDA, "FCDA")!;
-    const extRef = findElement(extRefXSWIPosDbPos, "ExtRef")!;
+    const fcda = findElement(orphanFCDA, "FCDA") as Element;
+    const extRef = findElement(extRefXSWIPosDbPos, "ExtRef") as Element;
 
     expect(doesFcdaMeetExtRefRestrictions(extRef, fcda)).to.be.false;
   });
 
   it("is restrictive with invalid ExtRef input", () => {
-    const fcda = findElement(orphanFCDA, "FCDA")!;
-    const extRef = findElement(invalidExtRef, "ExtRef")!;
+    const fcda = findElement(orphanFCDA, "FCDA") as Element;
+    const extRef = findElement(invalidExtRef, "ExtRef") as Element;
 
     expect(doesFcdaMeetExtRefRestrictions(extRef, fcda)).to.be.false;
   });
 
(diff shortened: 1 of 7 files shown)


tExtRef/doesFcdaMeetExtRefRestrictions.ts
```diff
@@ -26,10 +33,13 @@ export function doesFcdaMeetExtRefRestrictions(
 
   if (
     extRef.getAttribute("pServT") &&
-    controlBlockType !== extRef.getAttribute("pServT")
+    options?.controlBlockType !== extRef.getAttribute("pServT")
```
danyill, line 36:
```suggestion
    options.controlBlockType !== extRef.getAttribute("pServT")
```

I guess we don't need the question mark.

danyill: It seems a bit odd because the `controlBlockType` is not really optional.
If `pServT` is declared on the `extRef` we carry out this check and if `options.controlBlockType` is not declared I think we are checking "undefined" against the `pServT` value.

Does that sound right?

If it was really optional to check the control block type then that makes the `pServT` check also optional. In which case should the option be for the `pServT` check? Or should we be defaulting to not check the `pServT` if a control block isn't given (my preference).

In which case something like:

```js
if (
    extRef.getAttribute("pServT") && controlBlockType &&
    controlBlockType !== extRef.getAttribute("pServT")
    options?.controlBlockType !== extRef.getAttribute("pServT")
  )
```

I realise this hasn't functionally changed as part of this PR but I think it is worth discussing.

danyill (commented):
Thank you for providing this functionality and many apologies for the delay in reviewing.

I think the basic type functionality looks fine but I am a little concerned about the control block type check approach and maybe we can discuss a little more.

tExtRef/doesFcdaMeetExtRefRestrictions.ts
JakobVogelsang: So far it was very restrictive with missing control tag input. I think there are two ways to go:
1. Make `controlBlockType` not optional and force through typescript
2. Interpret missing `controlBlockType` definition as user does not want to check for `pServT`.

I have implemented the second now. Would love to heard which one you would prefer.

danyill: I am good with Option 2.
Thank you for considering this feedback :+1:

danyill (approved):

tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
```diff
@@ -14,96 +14,148 @@ const extRefXSWIBehQ = `<ExtRef pDO="Beh" pDA="q" pServT="GOOSE" pLN="CSWI"/>`;
...
+    const extRef = findElement(invalidExtRef1, "ExtRef") as Element;
 
     expect(doesFcdaMeetExtRefRestrictions(extRef, fcda)).to.be.true;
   });
 
-  it("is restrictive with missing controlServiceType", () => {
-    const fcda = findElement(publisherIED, `FCDA[desc="DPCdbpos"]`)!;
-    const extRef = findElement(extRefXSWIPosDbPos, "ExtRef")!;
+  it("does not check pServT with missing controlServiceType", () => {
+    const fcda = findElement(publisherIED, `FCDA[desc="DPCdbpos"]`) as Element;
```
danyill, line 42: `desc` is not valid on an FCDA -- we should probably avoid this...

danyill:
I guess it's OK if you don't want to fix the `desc` and schema validity for this -- it's a bit nitpickety :wink: 

image

danyill (approved):

JakobVogelsang:
> I guess it's OK if you don't want to fix the `desc` and schema validity for this -- it's a bit nitpickety 😉
> 
> image

Not in this PR I would not. I have added an issue that holds this thought.

PR #72, opened 2023-12-24
fix(tIED): Refactor insertIED to avoid duplicate ConnectedAPs

### LINKED ISSUE #71: tIED: insertIED duplicates ConnectedAPs
Currently, we duplicate ConnectedAPs during an Import under some circumstances.

https://github.com/OpenEnergyTools/scl-lib/blob/671b2fc11c7fece31a0fc22f38a05f3af78a6ffc/tIED/insertIED.ts#L9-L60

Consider the following file:

[XAT_BusA_P1_Template.zip](https://github.com/OpenEnergyTools/scl-lib/files/13746720/XAT_BusA_P1_Template.zip)

Which I am using with: https://danyill.github.io/oscd-import-ieds/index.deploy.html or https://danyill.github.io/scl-editor/

If I start with a new project and import all IEDs in this file I get a schema validation error.

The cause is this:

* When we import the first IED in the file we create the SubNetworks as we need them and we copy them with all their children.
* When we import the second IED in the file if the SubNetwork exists we just copy the ConnectedAPs.
* We don't check to see if an identical ConnectedAP already exists.

danyill:
Closes #71

Also, renames the `importIED` folder to `insertIED`, hope this is acceptable.

My formatter seems to lint differently but I am not sure why, sorry about this - I did use `npm run format`.

I'd be grateful for a review and happy to make improvements!


JakobVogelsang:
Thank you very much. This is a great contribution. I shall have a look.

To resolve the issue with the formatting, you need to `npm install`. Chris has added the prettier as a dependency in the packege.json.

danyill:
>To resolve the issue with the formatting, you need to npm install. Chris has added the prettier as a dependency in the packege.json.

Done, thanks for the tip.

> I shall have a look.

As always, gracias.

tIED/insertIED.ts
```diff
@@ -6,6 +6,23 @@ export type InsertIedOptions = {
   addCommunicationSection: boolean;
 };
 
+/**
+ * Copies an SCL SubNetwork element but without its ConnectedAP children.
+ * @param subNetwork - SCL SubNetwork element.
+ * @returns cloned SubNetwork without Element children.
+ */
+function getNewSubNetwork(subNetwork: Element) {
```
JakobVogelsang, line 14: The return type definition is missing, isn't it?

danyill: Quite right, fixed

JakobVogelsang (changes requested):
It seems to me that some of the formatting is not following with what we do in all the other files. @ca-d has added the dependancy `prettier` that does the formating. To use it make sure you have prettier installed on your machine by rebase on main and `npm install` and then run `npm run format`. Other than that the PR is ok.

danyill:
I've rebased on main and run `npm run format` which has also modified a few other files by default:

```
        modified:   package-lock.json
        modified:   tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
        modified:   tExtRef/doesFcdaMeetExtRefRestrictions.ts
        modified:   tExtRef/subscribe.ts
        modified:   tIED/insertIED.ts
```

So I've reverted all the ones except related to this PR.

JakobVogelsang (approved):
LDTM

Issue #79, opened 2024-01-20
tLN: Remove supervision with valid control block reference but no related subscriptions


danyill:
If a supervision exists but has no associated subscriptions on an IED I allow it to be overwritten. But our current scl-lib API does not allow this and so my test fails. 

The cause is:

https://github.com/OpenEnergyTools/scl-lib/blob/f67d73dd6c46a07ef0a19f1574f992c61358cc98/tLN/supervision/canInstantiateSubscriptionSupervision.ts#L152-L174

In the lower lines because there is a valid object reference I can't "remove" the supervision. I also don't have an option to force this. 

Should we have:

- a check on whether there are any ExtRefs subscribed (is it OK to supervise an unused control block?)
- a "force" option here?
- Or, once we fix up the remove subscription supervision (to only removal the Val text content, see #18) should I just remove the supervision first?

danyill:
I think an option called `replace` would be sufficient.

JakobVogelsang:
Dear @danyill. I took a deeper look into the issue and am afraid there is more work to be done. The function is now doing only insert, not remove. So replace is not possible out of the box. I feel the complexity in this function comes from the fact that we allow to have `LN` and `IED` as supervision sink. I would like to pull those apart and would suggest having an additional function like so

`function editSubscriptionSupervision(ln: Element, options: ...)`

It would be interesting to me what of the `SupervisionOptions` you would need for this function. 

What do you think about this approach?

danyill:
I'm not sure how to answer this at the moment :shrug:, we might need to discuss...

danyill:
> I feel the complexity in this function comes from the fact that we allow to have LN and IED as supervision sink. I would like to pull those apart and would suggest having an additional function like so
>
> function editSubscriptionSupervision(ln: Element, options: ...)
> 
> It would be interesting to me what of the SupervisionOptions you would need for this function.

I think pulling them apart is a good idea, the logic and different features required for the two cases was becoming excessively complex and difficult for end-users.

Regarding options, I think we should have:

* `checkDuplicateSupervisions` - which would disallow supervising the same control block within the same IED
* `checkSubscriptionExists` - which would disallow supervising a control block if the control block is not also subscripted onto an ExtRef

In my heart, I am unsure that `checkEditableSrcRef` is a required option - it seems to work around the intent of the standard and perhaps the best reason to retain is that the standard is not quite as clear as it could be. I'd most likely try to begin without it.

PR #81, opened 2024-01-27
Check supervision creation without passing a control block (closes #80).

### LINKED ISSUE #80: tLN: Check if supervision can be created without passing a control block
It would be generally useful to know if I can create supervisions without having to pass a control block.

As long as a supervision LN has been passed, this should be possible.

Although I worry `canInstantiateSubscriptionSupervision` is now quite complex, I'd like the control block to be an optional argument.

Then, as an end user if I want to check I also have to pass a supervision instance (I do not propose a rewrite but perhaps our API is not granular enough, I would prefer to just pass an IED and a type, e.g. 'LGOS' or 'LSVS'). But since the standard says this must exist I think it is OK.

This has two uses:

* An IED _could_ have LGOS and LSVS instances some of which are editable and others which are not. I wish to visually show which ones can and cannot be edited.
* I wish to evaluate if an IED can have supervisions edited without having to try to actually carry out supervision of a control block.

danyill:
Closes #80.

I don't know if this is even a good idea, and if it is not I am happy for this PR to be closed, it is intended to explore the issue.

To be useful this would also require #79.

### DIFF
--- a/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
+++ b/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
@@ -50,7 +50,7 @@ describe("Function compare FCDA basic types to ExtRef restrictions", () => {
     const extRef = findElement(extRefXSWIPosDbPos, "ExtRef") as Element;
 
     expect(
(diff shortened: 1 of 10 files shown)


danyill:
Let's close this for now - it's likely not quite right and it's getting stale.

PR #82, opened 2024-01-27
Allow replacement of supervision control block reference, closes #79.

### LINKED ISSUE #79: tLN: Remove supervision with valid control block reference but no related subscriptions

danyill:
Closes #79.

I also don't know if any of this is a good idea.

I've left a few formatting updates from `npm run format` in here.

### DIFF
--- a/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
+++ b/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
@@ -50,7 +50,7 @@ describe("Function compare FCDA basic types to ExtRef restrictions", () => {
     const extRef = findElement(extRefXSWIPosDbPos, "ExtRef") as Element;
 
     expect(
-      doesFcdaMeetExtRefRestrictions(extRef, fcda, { controlBlockType: "SMV" })
+      doesFcdaMeetExtRefRestrictions(extRef, fcda, { controlBlockType: "SMV" }),
     ).to.be.false;
   });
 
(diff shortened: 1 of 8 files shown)


JakobVogelsang:
Dear @danyill. I just had the time to look at your RP. Thank you for pushing the `scl-lib`. 

There are two option that I see. 
1. Replace be inherently set when the supervision.iedOrLn is `LN`. Then we just move the existing supervision to a new logical node. This required the iedOrLn be empty (no Val element with inner text)
2. Remove the supervision before creating a new one. For this could just export the `removeSubscriptionSupervision` function. Then still all check would apply and you would have some fail savety.

JakobVogelsang:
There are updates on #79. I would like to close this PR and finalize the refinement on #79 first.

PR #84, opened 2024-01-30
feat(tLN): removeSupervision

### LINKED ISSUE #18: (tLN) Remove subscription supervision

JakobVogelsang:
Closes #18 

I did create a new function as the name does more explain the function. Compared to `removeSubscriptionSupervision` here no supervision of a subscription is removed but rather the supervision directly. The two options allow to configure the function to individual needs.

@danyill I would be happy for a review.

### DIFF
--- a/tLN/supervision/removeSupervision.ts
+++ b/tLN/supervision/removeSupervision.ts
@@ -1,0 +1,68 @@
+import { Remove } from "../../foundation/utils.js";
+import { isSrcRefEditable } from "./foundation.js";
+
+export type removeSupervisionOptions = {
+  removeSupervisionLn: boolean;
+  checkSubscription: boolean;
+};
+
+/** @returns Whether other subscribed ExtRef of the same control block exist */
+function isControlBlockSubscribed(supervisionLn: Element): boolean {
+  const controlObjRef = supervisionLn.querySelector(
+    `:scope > DOI[name="GoCBRef"] > DAI[name="setSrcRef"] > Val, 
+    :scope > DOI[name="SvCBRef"] > DAI[name="setSrcRef"] > Val`
+  )?.textContent;
+
+  const extRefs = Array.from(
+    supervisionLn.closest("IED")!.querySelectorAll(
+      `:scope > AccessPoint > Server > LDevice > LN0 > Inputs > ExtRef, 
+      :scope > AccessPoint > Server > LDevice > LN0 > Inputs > ExtRef`
+    )
+  );
+
+  return extRefs.some((extRef) => {
+    const [srcCBName, srcLDInst, iedName] = [
+      "srcCBName",
+      "srcLDInst",
+      "iedName",
+    ].map((attr) => extRef.getAttribute(attr));
+
+    return controlObjRef === `${iedName}${srcLDInst}/LLN0.${srcCBName}`;
+  });
+}
+
+export function canRemoveSupervision(
+  supervisionLn: Element,
+  options: removeSupervisionOptions
+): boolean {
+  if (!isSrcRefEditable(supervisionLn)) return false;
+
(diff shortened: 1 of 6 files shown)


tLN/supervision/removeSupervision.ts
```diff
@@ -0,0 +1,72 @@
...
+    checkSubscription: true,
+  },
+): Remove | null {
+  if (!canRemoveSupervision(supervisionLn, options)) return null;
+
+  const val = supervisionLn.querySelector(
+    `:scope > DOI[name="GoCBRef"] > DAI[name="setSrcRef"] > Val, 
+    :scope > DOI[name="SvCBRef"] > DAI[name="setSrcRef"] > Val`,
+  );
+  if (!val) return null;
```
danyill, line 58: Should we re-order this. Even if there is no `Val` element I think I should be able to use `removeSupervisionLn`.

tLN/supervision/removeSupervision.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const canRemoveLn = supervisionLn.querySelector(
+    ':scope > Private[type="OpenSCD.create"]',
+  );
+
+  if (canRemoveLn && options.removeSupervisionLn)
+    return { node: supervisionLn };
```
danyill, line 65: I think we should not require the `OpenSCD.create` private before allowing removal.

It seems to me that IEC 61850-7-1 Ed 2.1 talks specifically about "capabilities description provided by the ICT in order that the SCT can add/delete or not LGOS/LSVS supervision logical nodes" in Annex G, section G.1 General.

I think this means if we pass the `canRemoveSupervision` test we can either change the Val  text content or add/remove supervision instances completely.

And if I'm wrong I think we should leave this up to the user and not restrict it to our private type.

tLN/supervision/removeSupervision.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export type removeSupervisionOptions = {
+  removeSupervisionLn: boolean;
+  checkSubscription: boolean;
+};
+
+/** @returns Whether other subscribed ExtRef of the same control block exist */
+function isControlBlockSubscribed(supervisionLn: Element): boolean {
+  const controlObjRef = supervisionLn.querySelector(
+    ':scope > DOI > DAI[name="setSrcRef"] > Val',
```
danyill, line 12: Oh my favourite! We must scope that DOI a little more tightly :stuck_out_tongue_winking_eye: to either:

`DOI[name="GoCBRef"]` or `DOI[name="SvCBRef"]`

danyill (changes requested):
Thank you for this and also for the opportunity to review this PR.

I think it would be good to ensure we could remove a supervision node whether or not there is a `Val` element and also that removing a supervision node should not depend on having a specific `Private`.

Happy to discuss!

tLN/supervision/removeSupervision.ts
danyill: Oh yes, and also this bit:

image

JakobVogelsang: I think I get now what you were about. I agree. I would want to change that as well in the other remove function and per default remove only the val element.

tLN/supervision/removeSupervision.ts
JakobVogelsang: So true :)

danyill (approved):
LGTM, thank you for incorporating this feedback.

PR #86, opened 2024-02-10
fix(tExtRef): allow LLN0 as default srcLNClass (closes #85).

### LINKED ISSUE #85: tExtRef: Match Source Attributes doesn't match srcLNClass correctly
This function doesn't appear to correctly match `srcLNClass`. If it is missing it should default to `LLN0`.

I think we need something like:

```js
 extRef.getAttribute("srcLNClass") ?? "LLN0" === srcLNClass
```

https://github.com/OpenEnergyTools/scl-lib/blob/7cbe03ae29c508af819ae57892b84ea73ca42345/tExtRef/matchSrcAttributes.ts#L8-L22

danyill:
Closes #85 but also a bit of a drive by improvement:

* Also allow `srcLNClass` and `srcLDInst` to be empty when choosing fallbacks.
* Allow `ldInst` to be used if `srcLDInst` is not defined.

This is as per Table 33 of IEC 61850-6 Ed 2.1.

I shall be happy to receive a review and make improvements.

### DIFF
--- a/tExtRef/matchSrcAttributes.spec.ts
+++ b/tExtRef/matchSrcAttributes.spec.ts
@@ -60,6 +60,45 @@ describe("matchSrcAttributes", () => {
     expect(matchSrcAttributes(extRef, gse)).to.equal(true);
   });
 
+  it("allows LN0 as parent when srcLNClass is undefined", () => {
+    extRef.removeAttribute("srcPrefix");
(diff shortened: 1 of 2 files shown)


JakobVogelsang (approved):
Wow thank you very much. Clear issue description, well defined naming and nice tests. LVGTM

PR #88, opened 2024-02-18
feat(tExtRef): add checkOnlyBType to subscribe options

### LINKED ISSUE #77: tExtRef: Add checkOnlyBType to subscribe
I was expecting a subscribe option to have an option for: "checkOnlyBType" and I think I missed that we didn't extend this all the way as a subscription option:

https://github.com/OpenEnergyTools/scl-lib/issues/65
https://github.com/OpenEnergyTools/scl-lib/pull/66

It seems we still need to add the option to the the subscribe options and pass through to doesFcdaMeetExtRefRestrictions so the API is exposed.

danyill:
Closes #77

I've just copied the preceding tests and added the `checkOnlyBType` option.

I'll be most happy to receive a review and make improvements.


JakobVogelsang (changes requested):
LGTM appart the doc for this option. I would love it to be consistent. That does not mean that the existing one is the better coice ;)

danyill:
> LGTM appart the doc for this option. I would love it to be consistent. That does not mean that the existing one is the better coice ;)

Sorry, I have not understood your concern - could you rephrase?

I've enhanced the type `SubscribeOptions` in line with the other options I think?

https://github.com/danyill/scl-lib/blob/6423b8bb75f8700306391a76a813a33c3428d8f7/tExtRef/subscribe.ts#L13-L20

JakobVogelsang:
Oh, sorry I was not clear I think. The same key had different comments in the different type they are used by. I think we could make the use of the library easier by equal descriptions for equal keys.

danyill:
I have made some updates, thanks :+1:

JakobVogelsang (approved):
LGTM

PR #92, opened 2024-03-29
feat(tIED): allow IED removal (closes #91).

### LINKED ISSUE #91: tIED: Remove IED
It would be helpful to me if we had an SCL lib function which allowed the removal of an IED.

Generally I think the same elements covered by https://github.com/OpenEnergyTools/scl-lib/blob/main/tIED/updateIED.ts need to be modified.

@JakobVogelsang are you happy for me to have a go at this?

```js
export function removeIED(ied: Element): Edit[] {

}
```

Should we be able to remove multiple IEDs or is an atomic IED removal sufficient (I think it is).

danyill:
Closes #91

As always, I would be very grateful for a review.

I'm not removing supervisions using the inbuild unsubscribe functionality because it's implemented "the wrong way" (removal of LNs or DOIs when I want removal of the GoCBRef/SvCBRef > Val's text content). I hope that's acceptable for now.

While I could mostly reuse the test structure for the `updateIED` code (thank you!), I have redone the test files as I needed the subscriptions and supervisions to be more realistic.

Once we have this right, I'd be keen for a new release as I have a menu plugin that I would like to provide in our distribution to allow IED removal.

### DIFF
--- a/tIED/removeIED.ts
+++ b/tIED/removeIED.ts
@@ -1,0 +1,95 @@
+import { isPublic } from "../tBaseElement/isPublic.js";
+import { unsubscribe } from "../tExtRef/unsubscribe.js";
+import { removeSubscriptionSupervision } from "../tLN/removeSubscriptionSupervision.js";
+
+import type { Remove, Update } from "../foundation/utils.js";
+
+const elementsToRemove = ["Association", "ClientLN", "ConnectedAP", "KDC"];
+
+const elementsToReplaceWithNone = ["LNode"];
+
+function removeIEDNameTextContent(ied: Element, iedName: string): Remove[] {
+  return Array.from(ied.ownerDocument.getElementsByTagName("IEDName"))
+    .filter(isPublic)
+    .filter((iedNameElement) => iedNameElement.textContent === iedName)
+    .map((iedNameElement) => {
+      return { node: iedNameElement };
+    });
+}
+
+function removeWithIedName(ied: Element, iedName: string): Remove[] {
+  const selector = elementsToRemove
+    .map((iedNameElement) => `${iedNameElement}[iedName="${iedName}"]`)
+    .join(",");
+
+  return Array.from(ied.ownerDocument.querySelectorAll(selector))
+    .filter(isPublic)
+    .map((element) => {
+      return { node: element };
+    });
+}
+
+function removeIedSubscriptionsAndSupervisions(
+  ied: Element,
+  iedName: string,
+): (Update | Remove)[] {
+  const extRefs = Array.from(ied.ownerDocument.querySelectorAll(":root > IED"))
+    .filter((ied) => ied.getAttribute("name") !== iedName)
+    .flatMap((ied) =>
+      Array.from(
+        ied.querySelectorAll(
+          `:scope > AccessPoint > Server > LDevice > LN0 > Inputs > ExtRef[iedName="${iedName}"], 
+            :scope > AccessPoint > Server > LDevice > LN > Inputs > ExtRef[iedName="${iedName}"]`,
+        ),
+      ),
+    );
+
+  const supervisionRemovals = removeSubscriptionSupervision(extRefs);
+  const extRefRemovals = unsubscribe(extRefs, { ignoreSupervision: true });
+
+  return [...extRefRemovals, ...supervisionRemovals];
+}
+
+function updateIedNameToNone(ied: Element, iedName: string): Update[] {
+  const selector = elementsToReplaceWithNone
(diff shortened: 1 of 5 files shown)


danyill:
Currently, this does not remove empty SubNetwork elements if we are removing the last ConnectedAP.

I guess we could do something like the following but I worry about over-reach. And I guess we could make it an option. Perhaps we could just wait until it's a problem for someone :wink: 

```js
function removeWithIedName(ied: Element, iedName: string): Remove[] {
  const selector = elementsToRemove
    .map((iedNameElement) => `${iedNameElement}[iedName="${iedName}"]`)
    .join(",");

  return Array.from(ied.ownerDocument.querySelectorAll(selector))
    .filter(isPublic)
    .map((element) => {
      if (
        element.tagName === "ConnectedAP" &&
        element.parentElement!.querySelectorAll("ConnectedAP").length === 1
      ) {
        return { node: <Element>element.parentElement };
      }
      return { node: element };
    });
}
```

tIED/removeIED.ts
```diff
@@ -0,0 +1,103 @@
...
+ * 2. Update LNodes to an iedName of None
+ * ```
+ * @param remove - IED element as a Remove edit
+ * @returns - Set of additional edits to relevant SCL elements
+ */
+export function removeIED(remove: Remove): (Update | Remove)[] {
+  if (
+    remove.node.nodeType !== Node.ELEMENT_NODE ||
+    remove.node.nodeName !== "IED" ||
+    !(<Element>remove.node).hasAttribute("name")
```
JakobVogelsang, line 89: Would you be so nice and change to `(remove.node as Element).hasAttribute("name")`. Just for consistency reasons.

tIED/removeIED.ts
```diff
@@ -0,0 +1,103 @@
...
+
+function updateIedNameToNone(ied: Element, iedName: string): Update[] {
+  const selector = elementsToReplaceWithNone
+    .map((iedNameElement) => `${iedNameElement}[iedName="${iedName}"]`)
+    .join(",");
+
+  return Array.from(ied.ownerDocument.querySelectorAll(selector))
+    .filter(isPublic)
+    .map((element) => {
+      return { element, attributes: { iedName: "None" } };
```
JakobVogelsang, line 69: Should the `ldInst` be set to null as well?

tIED/removeIED.ts
```diff
@@ -0,0 +1,103 @@
...
+    });
+}
+
+function removeIedSubscriptionsAndSupervisions(
+  ied: Element,
+  iedName: string
+): (Update | Remove)[] {
+  const extRefs = Array.from(ied.ownerDocument.querySelectorAll(":root > IED"))
+    .filter(
+      (ied) => ied.hasAttribute("name") && ied.getAttribute("name") !== iedName
```
JakobVogelsang, line 44: I think `ied.getAttribute("name") === iedName` should be enough. You know iedName is not null.

tIED/removeIED.ts
```diff
@@ -0,0 +1,103 @@
...
+import { removeSubscriptionSupervision } from "../tLN/removeSubscriptionSupervision.js";
+
+import type { Remove, Update } from "../foundation/utils.js";
+
+const elementsToRemove = [
+  "Association",
+  "ClientLN",
+  "ConnectedAP",
+  "KDC",
+  "IEDName",
```
JakobVogelsang, line 12: That has no effect, right. `IEDName` has IED.name in textContent

JakobVogelsang (changes requested):
That was a pleasure to review. Thank you. Well structure and nice tests. There are just very minor issue that I have spotted.

JakobVogelsang:
I think an issue marking the fact that the `removeIed` does not clean the `DataTypeTemplates` section would also be nice. I don't think we should have that from the beginning, but a reminder is nice to have.

tIED/removeIED.ts
danyill: I don't know. Couldn't that be used for specification? If I had a virtual IED and I was swapping virtual IEDs, might I want to retain the `ldInst`?

tIED/removeIED.ts
danyill: Quite correct.

tIED/removeIED.ts
danyill: Good spotting.

tIED/removeIED.ts
danyill: Well for now, let's remove it and see if it's a cause of trouble. It's a bit unclear to me what the valid combinations are here.

danyill:
Thank you for this review, I have made some updates.

PR #94, opened 2024-03-30
fix(tIED/insertIED): Ensure DTT collisions result in new unique IDs (closes #93)

### LINKED ISSUE #93: tIED: Insert IED does not correctly check when creating new datatype templates
I've created an `oscd-import-templates` which is just a simpe wrapper around our `insertIed` function in this library.

We found a case which is a little concerning:

1. Import some templates in such a way that templates need to be renamed to avoid a conflict.
2. When they are renamed the new one will be named using some nomenclature we made up for each datatype template:
    https://github.com/OpenEnergyTools/scl-lib/blob/647d18b3bd5d63265fcd936c9d4a4b2fd3164d75/tIED/insertIED.ts#L251-L263
3. This is transiently OK (on a single run of the function) but it is definitely not OK to run it a second time because we collide with our own uniqueness algorithm resulting in duplicate datatype templates.
4. What we need is a better uniqueness algorithm -- either to get a very low collision probability or (better) to check if it is used before deciding it is a suitable new name.
5. The following branch has a test case fail demonstrating this. In this case, two IEDs were imported and one (formerly `SIEMENS_7UT85_01` was renamed to `XAT_T1_P2`). The import order resulted in some templates being prefixed with `SIEMENS_7UT85_01`. When this tool was run again and identified conflicts it tried to generate the same ID to remove the collision (creating another collision which it doesn't resolve or identify).
    https://github.com/danyill/scl-lib/tree/issue-93-dtt-collision
7. I am happy to prepare a PR for this if we can agree on the way forward. Ideally I'd like to close this bug out in the next month or so.

See also: https://github.com/danyill/oscd-import-templates/issues/2

danyill:
Closes #93.

I'm not sure if this is a good way to solve this.

I have tried to add a regression test although it seems a bit complicated I think it represents the behaviour I am trying to avoid.

I will be delighted to receive a review and make improvements.

### DIFF
--- a/tIED/insertIED.ts
+++ b/tIED/insertIED.ts
@@ -134,6 +134,37 @@ function isDataTypeConnectionToIed(
     .some((anyLn) => anyLn.getAttribute("lnType") === id);
 }
 
+/**
+ * Generates a new DTT section id in case of a collision by adding a number in hex
+ * prefixed with an underscore to the existing id.
+ *
+ * @param existingId - Existing string for id attribute.
+ * @param existingLNodeType - Existing LNodeType for which a new ID is required.
+ * @param dataTypeTemplates - Existing Element of DTTs from the SCL file.
+ * @returns New string which is unique in the DTT section.
+ */
+function generateNewId(
+  existingId: string,
+  newIed: Element,
+  existingLNodeType: Element,
+  dataTypeTemplates: Element,
+): string {
+  const iedName = newIed.getAttribute("name");
+  const tagType = existingLNodeType.tagName;
+  const duplicatesFound = function (counter: number): boolean {
+    return !!dataTypeTemplates.querySelector(
+      `${tagType}[id="${existingId}@${iedName}#${counter.toString(10)}"]`,
+    );
+  };
+
+  let counter = 1;
+  // 2,000 limit reasonable way to avoid infinite loop
+  while (duplicatesFound(counter) && counter < 2000) {
+    counter += 1;
+  }
+  return `${existingId}@${iedName}#${counter.toString(10)}`;
+}
+
 function addEnumType(
   newIed: Element,
   newEnumType: Element,
@@ -150,8 +181,14 @@ function addEnumType(
     // There is an `id` conflict in the project that must be resolved by
     // concatenating the IED name with the id
     const data: Element = newEnumType.parentElement!;
-    const idOld = newEnumType.getAttribute("id");
-    const idNew = newIed.getAttribute("name")! + idOld;
+    const idOld = newEnumType.getAttribute("id")!;
+
+    const idNew = generateNewId(
+      idOld,
+      newIed,
+      existEnumType,
+      oldDataTypeTemplates,
+    );
     newEnumType.setAttribute("id", idNew);
 
     data
@@ -184,8 +221,14 @@ function addDAType(
     // There is an `id` conflict in the project that must be resolved by
     // concatenating the IED name with the id
     const data: Element | null = newDAType.parentElement!;
-    const idOld = newDAType.getAttribute("id");
-    const idNew = newIed.getAttribute("name")! + idOld;
+    const idOld = newDAType.getAttribute("id")!;
+
+    const idNew = generateNewId(
+      idOld,
+      newIed,
+      existDAType,
+      oldDataTypeTemplates,
+    );
     newDAType.setAttribute("id", idNew);
 
     data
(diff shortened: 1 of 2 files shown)


danyill:
I decided against counting in hex...

danyill:
Ooops, a little more testing shows all is not well. Lett me make that draft again.

tIED/insertIED.ts
```diff
@@ -134,6 +134,36 @@ function isDataTypeConnectionToIed(
...
+ * @returns New string which is unique in the DTT section.
+ */
+function generateNewId(
+  existingId: string,
+  existingLNodeType: Element,
+  dataTypeTemplates: Element,
+): string {
+  const tagType = existingLNodeType.tagName;
+  const duplicatesFound = function (id: string): boolean {
+    return dataTypeTemplates.querySelector(`${tagType}[id="${id}"]`) !== null;
```
danyill, line 153:
```suggestion
    return !dataTypeTemplates.querySelector(`${tagType}[id="${id}"]`);
```

tIED/insertIED.ts
```diff
@@ -134,6 +134,36 @@ function isDataTypeConnectionToIed(
...
+
+  let counter = 1;
+  let id = `${existingId}_${counter.toString(10)}`;
+
+  // 2,000 limit reasonable way to avoid DOS (?)
+  while (duplicatesFound(id) && counter < 2000) {
+    counter += 1;
+    id = `${existingId}_${counter.toString(10)}`;
+  }
+  return id;
```
JakobVogelsang, line 164: I would probably `return ${existingId}_${counter.toString(10)}`

JakobVogelsang: To avoid creating strings in worst case 2000 times

tIED/insertIED.ts
```diff
@@ -150,8 +180,9 @@ function addEnumType(
     // There is an `id` conflict in the project that must be resolved by
     // concatenating the IED name with the id
     const data: Element = newEnumType.parentElement!;
-    const idOld = newEnumType.getAttribute("id");
-    const idNew = newIed.getAttribute("name")! + idOld;
```
JakobVogelsang, line 154: I am not sure anymore, but am fairly confident that I have read in the standard that there are two proposed ways on how to solve a potentially duplicate id. One I remember was the use of the attribute `iedType` the other one from what I remember was to extend with the IED name. Have you seen that as well?

JakobVogelsang (commented):
Very well written and nice test. I have a comment that I still would like to clearify

tIED/insertIED.ts
danyill: You are quite correct, there is something like that. In Ed 2.1, section 9.5.1 of IEC 61850-6 we have:

> All types are uniquely identified by their type id. On generation of the system SCD file from
> IED ICD files, the LN type identifications may have to change to keep uniqueness across all
> IED definitions. To keep possible semantic information of the type names, it is recommended
> to generate a new LN type name by concatenating the IED name (which shall be unique within
> the file) with the old LNodeType name (which shall be unique at least per IED). If a LN type is
> generally valid for several IEDs of different type, then the iedType attribute shall be defined
> as an empty string. If it is important to keep the relation of the LNodeType to the IED type,
> then iedType should be set to the same value as the IED’s type attribute. Especially if an IED
> configurator needs the LNodeType contents back unchanged, it shall bind the LNodeType to
> the IED type by setting the iedType attribute identical to the IED’s type attribute.

I don't really think it's a very good proposal TBH. However I am happy to take instruction.

IED names can change and then there's this "leftover" information which is not useful. Most of the tools we use append some information to the end (usually I think `_1`, `_2` off the top of my head). This is also more useful when the DTTs are ordered and you "line them up more easily" and see that there is a "similar but different" DTT.

However arguably avoiding collisions _should_ be done in a more complex way based on this statement:

> Especially if an IED configurator needs the LNodeType contents back unchanged, it shall bind the LNodeType to the IED type by setting the iedType attribute identical to the IED’s type attribute.

I'm also not at all sure this is sane - any tool should have to look up the DTTs like it is an ordered catalog.
However again I am happy to take instruction, however this also has a potential problem in that conflicts cannot be resolved any more if IEDs so specify and then the SCT cannot do its job.

So we could make some changes but on balance I'd prefer not to :wink:

tIED/insertIED.ts
danyill: Thanks I shall refactor.

tIED/insertIED.ts
danyill: I should add that this type of thing is exactly what makes the standard hard to implement :-(
"It is recommmended <insert some difficulty requirement or implementation detail that causes many other problems>"!

danyill:
Thank you for this review, I have made some changes and happy to make further modifications

tIED/insertIED.ts
danyill: It's also really unclear where that recommendation starts and ends...

JakobVogelsang:
I have a suggestion:

oldId@iedName#counter

danyill:
Sure, let's do that. Thank you for this discussion.

danyill:
I have updated the PR as per your suggestion

PR #99, opened 2024-04-13
feat(tExtRef): Allow INT32/Enum mapping and ENS/ENC with INS/INC mappings

### LINKED ISSUE #98: tExtRef: Support Enum to INT32 mappings and ENx to INx mappings
IEC 61850-6 I.2.1.2 indicates that the following mappings are allowed:

* ENS <> INS
* ENC <> INC
* ENG <> ING

The change is only in the stVal otherwise the CDCs are identical, see IEC 61850-7-3 F.4.9.

Enumerations at 32-bit integers and so they can be mapped to INT32.
This means that the basic types "Enum" and "INT32" should always be compatible.

This means, I think that:

* For preconfigured bindings the above CDCs are compatible (not currently in our subscription checks)
* If we are only doing the basic type check then "Enum" and "INT32" are compatible.

The changes required is in `tExtRef/doesFcdaMeetExtRefRestrictions` and I think is something like:

```ts
/**
 * This function checks if restrictions of an `ExtRef` element given by
 * `pDO` and optionally by `pDA`, `pLN` and `pServT` are met by the FCDA/FCD
 * @param extRef - The `ExtRef` element to be checked against
 * @param data - The `FCDA` element to be checked
 * @param controlBlockType - The control block type to check back with `pServT`
 * @returns Whether the FCDA basic types meet the restrictions of the
 * ExtRef element
 */
export function doesFcdaMeetExtRefRestrictions(
  extRef: Element,
  fcda: Element,
  options: DoesFcdaMeetExtRefRestrictionsOptions = { checkOnlyBType: false }
): boolean {
  // Vendor does not provide data for the check so any FCDA meets restriction
  if (!extRef.hasAttribute("pDO")) return true;

  const fcdaTypes = fcdaBaseTypes(fcda);
  const extRefSpec = extRefTypeRestrictions(extRef);

  // Check cannot be performed assume restriction check to fail
  if (!extRefSpec || !fcdaTypes) return false;

  // If service type specified it must match
  if (
    extRef.getAttribute("pServT") &&
    options.controlBlockType &&
    options.controlBlockType !== extRef.getAttribute("pServT")
  )
    return false;

  // Some vendors allow subscribing of e.g. ACT to SPS, both bType BOOLEAN
  if (options.checkOnlyBType) return fcdaTypes.bType === extRefSpec.bType;

  if (
    extRef.getAttribute("pLN") &&
    extRef.getAttribute("pLN") !== fcda.getAttribute("lnClass")
  )
    return false;

  // Ed 1 to Ed 2 compatibility. The following are compatible:
  // ENS -> INS, ENC -> INC and ENG -> ING
  if (
    fcdaTypes.cdc !== extRefSpec.cdc &&
    !(fcdaTypes.cdc === "ENS" && extRefSpec.cdc === "INS") &&
    !(fcdaTypes.cdc === "INS" && extRefSpec.cdc === "ENS") &&
    !(fcdaTypes.cdc === "ENC" && extRefSpec.cdc === "INC") &&
    !(fcdaTypes.cdc === "INC" && extRefSpec.cdc === "ENC") &&
    !(fcdaTypes.cdc === "ENG" && extRefSpec.cdc === "ING") &&
    !(fcdaTypes.cdc === "ING" && extRefSpec.cdc === "ENG")
  )
    return false;

  if (
    extRef.getAttribute("pDA") &&
    // Basic types not the same
    fcdaTypes.bType != extRefSpec.bType &&
    // Enum to INT32 mappings specifically allowed, Ed 1 to Ed 2 compatibility
    !(fcdaTypes.bType === "INT32" && extRefSpec.bType === "Enum") &&
    !(fcdaTypes.bType === "Enum" && extRefSpec.bType === "INT32")
  )
    return false;

  return true;
}
```

I will prepare a PR and some tests.

See https://github.com/danyill/oscd-subscriber-later-binding/issues/52

danyill:
Closes #98

I will be very pleased to receive a review and make improvements.

Once this is merged if we could have a release of scl-lib I would be grateful.

### DIFF
--- a/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
+++ b/tExtRef/doesFcdaMeetExtRefRestrictions.spec.ts
@@ -15,6 +15,10 @@ const invalidExtRef = `<ExtRef pDO="someInvalidDO" pDA="Dbpos" pServT="GOOSE" pL
 const invalidExtRef1 = `<ExtRef pDA="Dbpos" pServT="GOOSE" pLN="XSWI"/>`;
 const extRefCMVFLOAT32 = `<ExtRef pDO="A.phsA" pDA="cVal.mag.f" pServT="Report" pLN="MMXU"/>`;
 const extRefBlkOpnstVal = `<ExtRef pDO="BlkOpn" pDA="stVal" pLN="CILO" />`;
+const extRefENS = `<ExtRef pDO="Beh" pDA="stVal" />`;
+const extRefINS = `<ExtRef pDO="IntIn" pDA="stVal" />`;
+const extRefINC = `<ExtRef pDO="OpCntRs" pDA="stVal" />`;
+const extRefENC = `<ExtRef pDO="Mod" pDA="stVal" />`;
 
 describe("Function compare FCDA basic types to ExtRef restrictions", () => {
   it("is restrictive with invalid FCDA input", () => {
@@ -115,7 +119,7 @@ describe("Function compare FCDA basic types to ExtRef restrictions", () => {
     ).to.be.false;
   });
 
-  it("returns true for matching types CMV/FLOAT32 ", () => {
+  it("returns true for matching types CMV/FLOAT32", () => {
     const fcda = findElement(
       publisherIED,
       `FCDA[desc="CMVFLOAT32"]`,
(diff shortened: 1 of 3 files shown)


JakobVogelsang (approved):
LGTM

danyill:
Thanks for the review

PR #104, opened 2024-07-22
Fix conv rev update

### LINKED ISSUE #102: confRev updated incorrectly on SampledValue Control Block
We think the `confRev` is not being correctly updated on some control block types.

For instance in SV control blocks, an update of the `smvID` should not result in a change to `confRev`:

image

However we update this on almost any change:

https://github.com/OpenEnergyTools/scl-lib/blob/bd020d3a5534ec9da2915c5f1c7ff5518214b9d6/tSampledValueControl/updateSampledValueControl.ts#L9-L78

Unless there is a TISSUE clarifying this, I think this may need to be changed for GOOSE, SV and maybe also ReportControl blocks.

I'll be happy to prepare a PR.

JakobVogelsang:
Closes #102

### DIFF
--- a/tGSEControl/updateGSEControl.ts
+++ b/tGSEControl/updateGSEControl.ts
@@ -9,11 +9,11 @@ import { updateDatSet } from "../tControl/updateDatSet.js";
 /**
  * Utility function to update GSEControl element attributes.
  * ```md
- * These attributes trigger addition edits
+ * These attributes trigger additional edits such as
  * - name: also updates GSE.cbName and supervision references
  * - datSet: update reference DataSet.name - when DataSet is single use
  *
- * >NOTE: confRev attribute is updated +10000 every time this function is called
+ * >NOTE: confRev attribute is updated +10000 on every data set change
  * ```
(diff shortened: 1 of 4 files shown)


tGSEControl/updateGSEControl.ts
```diff
@@ -9,11 +9,11 @@ import { updateDatSet } from "../tControl/updateDatSet.js";
 /**
  * Utility function to update GSEControl element attributes.
  * ```md
- * These attributes trigger addition edits
+ * These attributes trigger additional edits such as
  * - name: also updates GSE.cbName and supervision references
  * - datSet: update reference DataSet.name - when DataSet is single use
  *
- * >NOTE: confRev attribute is updated +10000 every time this function is called
+ * >NOTE: confRev attribute is updated +10000 on every data set change
```
danyill, line 16: We should also make this change in the docstring for `updateSampledValueControl`

danyill (commented):
LGTM.

Do you think we should also address `updateReportControl` in this PR which I believe has the same issue?

PR #105, opened 2024-07-22
Fix udapte conf rev on remove

### LINKED ISSUE #103: Update confRev on dataset removal from control block
What we do is on removal of a dataset we sent `confRev` to 0.

https://github.com/OpenEnergyTools/scl-lib/blob/bd020d3a5534ec9da2915c5f1c7ff5518214b9d6/tDataSet/removeDataSet.ts#L14-L34

But I think it should be incremented instead, "inclusive removal of this" in the standard:

image

(Background https://github.com/OpenEnergyTools/oscd-publisher/issues/17)

JakobVogelsang:
Closes #103


danyill (approved):
LGTM, thanks!

PR #122, opened 2024-12-07
fix: updateIED IEDName and supervision

JakobVogelsang:
Is fixing https://github.com/OpenEnergyTools/open-scd-core/issues/18


danyill (approved):
LGTM, thank you for making this change. 

I would be pleased for an scl-lib release and then I'll re-introduce the new core back into our distribution

PR #128, opened 2025-02-09
feat(tIED): update internal object references on IED rename 

### LINKED ISSUE #125: tIED: updateIED does not update object references
I've been testing the rename IEDs function with some ICTs.

I believe we don't update the `setSrcRef` with rename IEDs in general.

```xml
<DOI name="InRef1">
    <DAI name="setSrcRef">
        <Val>XAT_232_P1PRO/DC10CILO10.EnaOpn.stVal</Val>
    </DAI>
</DOI>
```

Should we be generally processing a file's object references on a rename?

Perhaps this setSrcRef may be a bug in my CID file (it was hand crafted) and there shouldn't be IED references within the `Val`

image

danyill:
Closes #125

Not quite implemented as I thought on the issue but I hope this is better.

### DIFF
--- a/tIED/updateIED.spec.ts
+++ b/tIED/updateIED.spec.ts
@@ -132,4 +132,47 @@ describe("Function to an update the IED name attributes and its referenced eleme
     ).filter((iedName) => iedName.textContent?.startsWith("newIedName"));
     expect(after.length).to.equal(4);
   });
+
+  it("updates object references without checking permissions", () => {
+    const sclDom = new DOMParser().parseFromString(scl, "application/xml");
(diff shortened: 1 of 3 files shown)


danyill:
I've tested this by successfully loading a file on a device after doing a rename. Previously this didn't work with SEL Architect and it at least loads successfully on the device now and publishes GOOSE messages so it looks like the parser is happy.

I would be most grateful for a detailed review and happy to make improvements.

My implementation is designed to short-circuit as quickly as possible and to be fault tolerant. I think the code coverage is looking good.

Once we agree on this PR I would be most pleased if we could also do:

* An updated deployment of scl-wizarding
* A new release of scl-lib

JakobVogelsang (approved):

PR #130, opened 2025-02-25
fix(tLN): Make supervision removal only remove Val textContent

### LINKED ISSUE #129: tLN: Remove subscription supervision removes too much
We've recently found that for some manufacturer's IEDs (SEL on the SEL-411L-2) our default supervision removal is "damaging" and prevents the ICT tool from correctly opening the file.

What we do is remove the `DOI[name="GoCBRef"]`

https://github.com/OpenEnergyTools/scl-lib/blob/0ce54dfcb035a599262cd707402b823b1ab427ba/tLN/removeSubscriptionSupervision.ts#L121-L137s

The ICT software then complains that it cannot find this DOI and refuses to open the overall SCD file.

I'd like to propose that we narrow the scope so that "subscription removal means only removal of the text content on the Val element".

My reading of IEC 61850-7-1 Ed 2.1 is that the SCT may remove either the complete LN or just the content of the Val element.

Image

danyill:
Closes #129 

This PR makes supervision removal do the most minimal thing possible and seems to me to be the best way of reading the standard. It also fixes an issue where an ICT tool cannot function with our present behaviour.

### DIFF
--- a/tLN/removeSubscriptionSupervision.ts
+++ b/tLN/removeSubscriptionSupervision.ts
@@ -14,25 +14,29 @@ type GroupedExtRefs = {
 function removableSupervisionElement(
   ctrlBlock: Element,
   subscriberIed: Element,
-): Element | null {
+): Node | Element | null {
   const supervisionType = ctrlBlock.tagName === "GSEControl" ? "LGOS" : "LSVS";
+  const doiName = ctrlBlock.tagName === "GSEControl" ? "GoCBRef" : "SvCBRef";
 
   const valElement = Array.from(
     subscriberIed.querySelectorAll(
-      `LN[lnClass="${supervisionType}"] > DOI > DAI > Val`,
+      `LN[lnClass="${supervisionType}"] > DOI[name="${doiName}"] > DAI[name="setSrcRef"] > Val`,
     ),
   ).find((val) => val.textContent === controlBlockObjRef(ctrlBlock));
   if (!valElement) return null;
 
   const ln = valElement.closest("LN")!;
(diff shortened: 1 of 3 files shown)


danyill:
I could likely do more to tidy up the tests here, let me know if that is of concern and I will update.

JakobVogelsang:
I would still like to have the ln removed if being created by the subscribe function. So if the private tag is in remove the ln otherwise remove the val.taxtContent. That would as far as I understand solve the issue #129, right?

danyill:
Yes, let's do that, thank you for the prompt feedback 👍

tLN/removeSubscriptionSupervision.ts
```diff
@@ -14,25 +14,29 @@ type GroupedExtRefs = {
...
   subscriberIed: Element,
-): Element | null {
+): Node | Element | null {
   const supervisionType = ctrlBlock.tagName === "GSEControl" ? "LGOS" : "LSVS";
+  const doiName = ctrlBlock.tagName === "GSEControl" ? "GoCBRef" : "SvCBRef";
 
   const valElement = Array.from(
     subscriberIed.querySelectorAll(
-      `LN[lnClass="${supervisionType}"] > DOI > DAI > Val`,
+      `LN[lnClass="${supervisionType}"] > DOI[name="${doiName}"] > DAI[name="setSrcRef"] > Val`,
```
danyill, line 23: Eek it was actually broken anyways!

danyill:
Thank you for the feedback, I have updated the PR as per your request and would, as always be grateful for a review and happy to make changes.

PR #146, opened 2026-04-27
Feat remove ied duplicates issue

### LINKED ISSUE #121: tIED: removeIeds behaviours with LNodes
Currently I find that the behaviour of `removeIED` for LNodes is to replace the IED name with `None` when the IED is removed.

This behaviour is problematic for me when removing a number of IEDs in the Remove IEDs plugin in Transpower's distribution.

It leaves a large number of dangling LNode elements.

For various reasons we often map an LPHD node for each IED against a busbar bay.

After removing several IEDs this section looks like this:

```xml
	<Function name="BusPhysConnection">
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="XAT_242_C1" ldInst="CFG" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="None" prefix="DevID" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" prefix="DevID" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" prefix="DevID" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" prefix="DevID" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="1"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
		<LNode iedName="None" lnClass="LPHD" lnInst="0"/>
	</Function>
```

This however becomes a schema violation:

image

The reason for this is spelt out in Table 8 of IEC 61850 Ed 2.1 and the subsequent text:

> For specifications where iedName=”None” everywhere, the combination of the other attributes
> must be unique within the same level. This means e.g. that the prefix or lnInst should be
> different if several LNs with identical lnClass are used within the same substation part (i.e.
> same bay). This should also be the case for IED names, if application related naming (see
> 8.5.4) shall be used. This is NOT checked by the SCL schema, therefore it is the
> responsibility of the project engineer or system tool, if application related naming shall be
> used additionally to product (IED)-related naming.

What should the behaviour of the `removeIED` function be here?

In my view at the least we need to provide an option to allow removal of LNodes to allow a bulk delete of IEDs without dangling, unnecessary schema non-compliance which is at least one common workflow option. I am confident it is also a valid workflow to sometimes retain the LNodes so I think it is a good case for an optional argument.

@JakobVogelsang WDYT?

Would you accept:

```ts
export type RemoveIedOptions = {
  removeLNodes: boolean;
};

function removeIED(remove: Remove, 
                   options: RemoveIedOptions = { removeLNodes: false }): (Update | Remove)[] {
 ... 
}
```

If you are happy  with this call signature I can prepare a PR.

stee-re:
Resolves #121
## Summary
 
When multiple IEDs reference the same LNode, removing those IEDs sequentially can result in duplicate LNode 
entries with `iedName="None"`. This PR adds an option to detect and remove such duplicates instead of 
creating redundant entries, by default. Callers can use the options to prevent this happening:

```typescript
remove(myIEDElement, {removeLNodes: false});
```
 
 ## Changes
 - **`tIED/removeIED.ts`**: 
   - Added `RemoveIedOptions` interface with a `removeLNodes` flag (defaults to `true`)
   - Introduced a `lNodeKey()` helper to compute a unique key per LNode
   - When `removeLNodes` is `true` (default), duplicate LNodes already set to `"None"` are removed rather 
than duplicated
   - When `removeLNodes` is `false`, LNodes are only updated to `"None"` (previous behaviour)
 - **`index.ts`**: Exported `RemoveIedOptions` type
 - **`tIED/removeIED.spec.ts`**: Added tests for duplicate removal and opt-out behaviour
 - **`tIED/removeIED.testfile.ts`**: Added test fixture with duplicate LNode scenario
 - **`.gitignore`**: Added IDE and build artifact patterns


tIED/removeIED.ts
```diff
@@ -50,31 +50,63 @@ function removeIedSubscriptionsAndSupervisions(
...
 
-function updateIedNameToNone(ied: Element, iedName: string): SetAttributes[] {
+function lNodeKey(element: Element): string {
+  const attr = (name: string) => element.getAttribute(name) ?? "";
+  return `${attr("lnInst")}|${attr("lnClass")}|${attr("ldInst")}|${attr(
+    "prefix",
+  )}`;
+}
+
+function updateIedNameToNone(
```
danyill, line 60: Should we call this something else? e.g. `virtualiseIedOrRemoveLNode` (too long) or `ensureUniqueLNodes` (doesn't quite capture it) or just `updateLNodes` (even if it can also remove them!).

Eek it's a bit of a mouthful but I think it's a bit misleading as is.

tIED/removeIED.ts
```diff
@@ -50,31 +50,63 @@ function removeIedSubscriptionsAndSupervisions(
   return [...extRefRemovals, ...supervisionRemovals];
 }
 
-function updateIedNameToNone(ied: Element, iedName: string): SetAttributes[] {
+function lNodeKey(element: Element): string {
+  const attr = (name: string) => element.getAttribute(name) ?? "";
+  return `${attr("lnInst")}|${attr("lnClass")}|${attr("ldInst")}|${attr(
+    "prefix",
+  )}`;
```
danyill: It is nit-picking but I suppose I'd prefer something a little more semantic:

```ts
return `${attr("ldInst")}|${attr("prefix")}|${attr("lnClass")}|${attr("lnInst")}`;
```

tIED/removeIED.ts
```diff
@@ -50,31 +50,63 @@ function removeIedSubscriptionsAndSupervisions(
...
+        if (parent) {
+          const key = lNodeKey(element);
+          const hasDuplicate = Array.from(
+            parent.querySelectorAll(':scope > LNode[iedName="None"]'),
+          ).some((sibling) => lNodeKey(sibling) === key);
+
+          if (hasDuplicate) return { node: element } as Remove;
+        }
+      }
+      return { element, attributes: { iedName: "None" } } as SetAttributes;
```
danyill, line 83: I suppose I don't like that we have an API that can needlessly result in a schema invalid file. I think if we agree we should have only two options, "virtualise" or "remove LNodes" from the recent issue discussions then we should change this API.

How do you feel about making this small modification such that if we're not virtualising we just Remove the LNode rather than doing a SettAttributes?

```suggestion
      return { node: element } as Remove;
```

I think then our options should reflect this terminology.

tIED/removeIED.ts
```diff
@@ -50,31 +50,63 @@ function removeIedSubscriptionsAndSupervisions(
...
+
 /**
  * Function to remove an IED.
  * ```md
  * The function makes sure to also:
  * 1. Remove all elements which should no longer exist including ClientLN,
  *    KDC, Association, ConnectedAP and IEDName
  * 2. Remove subscriptions and supervisions
- * 2. Update LNodes to an iedName of None
+ * 3. Update LNodes to an iedName of None (or remove duplicates)
```
danyill, line 100: Perhaps 

```suggestion
 * 3. Update LNodes to an iedName of None (and remove any duplicates by default)
```

danyill (changes requested):
Thank you for this contribution. It looks good to me but I wonder if it would be better if we didn't have an option to make the file schema invalid following ca-d comments in https://github.com/OpenEnergyTools/scl-lib/issues/121#issuecomment-4327484686

tIED/removeIED.ts
stee-re: done

stee-re:
@danyill I'm a little confuddled, this feedback appears to have come _after_ your comment here:
https://github.com/openscd/scl-lib/issues/121#issuecomment-4359122334

I like the function name "detachLNodeBindings()", it seems clear, or unbindLNodes(). 

Based on all the feedback from the scl-lib trinity, I will do the following:

- removeIED, by default, removes ALL Substation LNodes bound to the IED.
- removeIED supports the option preserveLNodes, which, if set to true, will use your function to set bound LNodes to "None" if one doesn't already exist - removes LNodes which would result in dups.

Please holler if I've missed something

PR #147, opened 2026-05-27
fix: removeIED - prevent duplicate LNode's

### LINKED ISSUE #121: tIED: removeIeds behaviours with LNodes

stee-re:
resolves #121
# Summary
When multiple IEDs reference the same LNode, removing those IEDs sequentially can result in duplicate LNode
entries with iedName="None". This PR modifies the default behaviour for `removeIED()` to delete all LNodes bound to that IED by default, but exposes a new optional flag `preserveLNodes` which will instead, set the LNodes ideName="None". In cases where setting the iedName="None" would lead to a duplicate LNode, then it is removed instead. 
E.g.
```typescript
remove(myIEDElement, {preserveLNodes: true});
```

# Changes
- **`tIED/removeIED.ts:`**
    - Added RemoveIedOptions interface with an optional preserveLNodes flag (defaults to false)
    - Added the following LNode related functions:
    - **`removeBoundLNodes()`** -  Default handling for LNodes - find any (public) matching LNodes and create a Remove edit for them.
    - **`detachLNodeBindings()`** - if "preserveLNodes flag set, this function is called instead, to instead set the LNodes iedName to "None", taking care to avoid duplicates. Any LNodes encountered which would cause a duplicate, are removed instead.

## Helper functions
- **`lNodeKey`** - creates a key for the LNode based on all the other attributes (except iedName) which combined, must be unique. 
- **`removeLNode()`** & **`setLNodeToNone()`** - oneliner functions, there only to make the code more readable. 
- `getLNodeByIedName()` used to find the bound LNodes (and the specification LNodes (set ot None)) 

# Other Changed files
 - **`index.ts`**: Exported `RemoveIedOptions` type
 - **`tIED/removeIED.spec.ts`**: Added tests for LNode removal and opt-in LNode updating behaviour
 - **`tIED/removeIED.testfile.ts`**: Added test fixture with duplicate LNode scenario
 - **`.gitignore`**: Added IDE and build artifact patterns
 
Code coverage 99.52%

### DIFF
--- a/tIED/removeIED.spec.ts
+++ b/tIED/removeIED.spec.ts
@@ -134,8 +134,7 @@ describe("Function to an remove the IED and its referenced elements", () => {
      * - Find all LNode references with matching iedName and either set them to None, or delete them if setting them to None would result in duplicate LNode keys within the same scope.
      *   The scope is defined as the nearest Bay, VL or Substation parent.
      */
-    describe("without 'preservveNodes' set (default)", () => {
-      //TODO consider changing this into a forEach (Substation, VL and Bay) array test.
+    describe("without 'preserveNodes' set (default)", () => {
       ["Bay", "VoltageLevel", "Substation"].forEach((scope) => {
         it(`deletes all LNodes found directly within a ${scope}`, () => {
           const sclDom = new DOMParser().parseFromString(
@@ -159,7 +158,8 @@ describe("Function to an remove the IED and its referenced elements", () => {
           // The number of LNodes set to None should not have changed.
           expect(after_spec_LNodeCount).to.equal(beforeSpec_LNodeCount);
 
-          //
+          const privateLNodeCount = sclDom.querySelectorAll(`${scope} > Private > LNode[iedName='IED_A']`).length;
+          expect(privateLNodeCount).to.equal(1, 'Private LNode count is wrong');
         });
       });
     });
@@ -169,7 +169,7 @@ describe("Function to an remove the IED and its referenced elements", () => {
       // Do keep in mind however, the subject SCL has 2 of everything. E.g. S1 & S2
       ["Bay", "VoltageLevel", "Substation"].forEach((scope) => {
         it(`Within a ${scope}, it sets all bound LNodes to None`, () => {
-          //we're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur (yet).
+          //We're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur (yet).
           const sclDom = new DOMParser().parseFromString(
             sclDuplicateLNodes,
             "application/xml",
@@ -196,7 +196,7 @@ describe("Function to an remove the IED and its referenced elements", () => {
 
       ["Bay", "VoltageLevel", "Substation"].forEach((scope) => {
         it(`Within a ${scope}, it removes 'would-be' duplicates`, () => {
-          //we're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur.
+          //We're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur.
           const sclDom = new DOMParser().parseFromString(
             sclDuplicateLNodes,
             "application/xml",
--- a/tIED/removeIED.spec.ts
+++ b/tIED/removeIED.spec.ts
@@ -5,7 +5,7 @@ import { isRemove, isUpdate } from "@openscd/oscd-api/utils.js";
 
 import { handleEdit } from "../foundation/helpers.test.js";
 
-import { scl } from "./removeIED.testfile.js";
+import { scl, sclDuplicateLNodes } from "./removeIED.testfile.js";
 
 import { removeIED } from "./removeIED.js";
 
@@ -47,10 +47,10 @@ describe("Function to an remove the IED and its referenced elements", () => {
     expect(removeIED({ node: publi }).length).to.equal(0);
   });
 
-  it("updates LNode iedName attributes to None as well", () => {
+  it("removes all bound LNodes", () => {
     const edits = removeIED({ node: subscriber1 });
 
-    expect(numberUpdates(edits, "LNode")).to.equal(1);
+    expect(numberRemoves(edits, "LNode")).to.equal(1);
   });
 
   it("removes ConnectedAPs as well", () => {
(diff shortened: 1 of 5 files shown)


tIED/removeIED.spec.ts
```diff
@@ -126,4 +126,136 @@ describe("Function to an remove the IED and its referenced elements", () => {
...
   });
+
+  describe("referenced LNode's", () => {
+    /*
+     * Here we need to test:
+     * - Delete all LNode references found with matching iedName, BUT only inside the substation section (but not inside Private sections).
+     * - Find all LNode references with matching iedName and either set them to None, or delete them if setting them to None would result in duplicate LNode keys within the same scope.
+     *   The scope is defined as the nearest Bay, VL or Substation parent.
+     */
+    describe("without 'preservveNodes' set (default)", () => {
```
danyill:
```suggestion
    describe("without 'preserveNodes' set (default)", () => {
```

tIED/removeIED.spec.ts
```diff
@@ -126,4 +126,136 @@ describe("Function to an remove the IED and its referenced elements", () => {
...
+      //TODO consider changing this into a forEach (Substation, VL and Bay) array test.
```
danyill:
```suggestion
```
It looks like an array already...

tIED/removeIED.spec.ts
```diff
@@ -126,4 +126,136 @@ describe("Function to an remove the IED and its referenced elements", () => {
...
+            sclDom.querySelectorAll(`${scope} LNode[iedName="IED_A"]`),
+          ).length;
+          const after_spec_LNodeCount = (
+            sclDom.querySelectorAll(`${scope} > LNode[iedName='None']`) ?? []
+          ).length;
+          expect(after_iedA_lNodes).to.equal(0);
+          // The number of LNodes set to None should not have changed.
+          expect(after_spec_LNodeCount).to.equal(beforeSpec_LNodeCount);
+
+          //
```
danyill:
```suggestion
```

Just removing a little crud.

tIED/removeIED.spec.ts
```diff
@@ -126,4 +126,136 @@ describe("Function to an remove the IED and its referenced elements", () => {
...
+        });
+      });
+    });
+
+    describe.only("with preserveLNodes set", () => {
+      // Broke this into 3 separate tests, so the scope of the failure "might" be narrower.
+      // Do keep in mind however, the subject SCL has 2 of everything. E.g. S1 & S2
+      ["Bay", "VoltageLevel", "Substation"].forEach((scope) => {
+        it(`Within a ${scope}, it sets all bound LNodes to None`, () => {
+          //we're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur (yet).
```
danyill:
```suggestion
          // We're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur (yet).
```

tIED/removeIED.spec.ts
```diff
@@ -126,4 +126,136 @@ describe("Function to an remove the IED and its referenced elements", () => {
     // 1 supervised control block is not subscribed so is not removed
     expect(after.length).to.equal(1);
   });
+
+  describe("referenced LNode's", () => {
+    /*
+     * Here we need to test:
+     * - Delete all LNode references found with matching iedName, BUT only inside the substation section (but not inside Private sections).
```
danyill, line 133: I thought after reading this you would have put an `LNode` inside a `Private` section to test this...

tIED/removeIED.spec.ts
```diff
@@ -126,4 +126,136 @@ describe("Function to an remove the IED and its referenced elements", () => {
...
+          expect(lNodes.length).to.equal(
+            beforeSpec_LNodeCount + beforeIedA_LNodeCount,
+          );
+          //
+        });
+      });
+
+      ["Bay", "VoltageLevel", "Substation"].forEach((scope) => {
+        it(`Within a ${scope}, it removes 'would-be' duplicates`, () => {
+          //we're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur.
```
danyill:
```suggestion
          // We're using the "duplicates" test file, but by only deleting 1 IED, no duplicates occur.
```

tIED/removeIED.ts
```diff
@@ -50,16 +48,94 @@ function removeIedSubscriptionsAndSupervisions(
...
+    doc.querySelectorAll(`Substation LNode[iedName=${name}]`) ?? [],
+  ).filter(isPublic);
+};
+
+/**
+ * Default handling for LNodes - find any (public) matching LNodes and create a Remove edit for them.
+ */
+function removeBoundLNodes(ied: Element, name: string): Remove[] {
+  return (getLNodesByIedName(ied.ownerDocument, name) ?? []).map(removeLNode);
+}
```
danyill, line 80: Is it me or is this very verbose? Or is that the AI secret sauce 😉 

Why not just?

```suggestion
function removeBoundLNodes(ied: Element, name: string): Remove[] {
  return (getLNodesByIedName(ied.ownerDocument, name) ?? []).map( { node: ln };);
}
```

tIED/removeIED.ts
```diff
@@ -50,16 +48,94 @@ function removeIedSubscriptionsAndSupervisions(
...
+  name: string,
+): (SetAttributes | Remove)[] {
+  const doc = ied.ownerDocument;
+  const boundNodes = getLNodesByIedName(doc, name);
+
+  if (boundNodes.length === 0) {
+    return [];
+  }
+
+  const UnboundLNodesByScope = new Map<Element, Set<string>>();
```
danyill:
```suggestion
  const unboundLNodesByScope = new Map<Element, Set<string>>();
```

Why would this start with a capital letter?

danyill (dismissed):
Thank you for looking at this and my apologies for the delay in reviewing.

I've flagged just a few small "tidyness", feel free to fix or not as you wish.

tIED/removeIED.ts
stee-re: Admittedly its not perfect, but I've been preferring the use of a descriptive function name when creating overloaded objects like this. As in, there are a few different kinds of Edits. This makes it abundantly clear which one it is. Granted, there were other signs :-) but of lately that has been my approach not an AI anomaly - worse, its a Stee-anomaly!!!

stee-re:
Thanks for your feedback - I couldn't believe how many typo's were in there!! Everyhing you pointed out, was addressed. Need a re-approve when you have a minute

danyill (approved):
LGTM, thank you for the explanations and updates!

danyill:
I merged a little too soon. 

1. I noticed we've reduced our test coverage slightly and I'd like this library to get as close to 100% test coverage as possible.
2. We have a `describe.only` in our tests so not all our tests are running within `removeIED.spec.ts`!
3. When the `describe.only` is replaced with `describe` we have test failures which need to be fixed.

I'll look into these and fix.

PR #149, opened 2026-07-12
fix: improve test coverage, remove scope guards

danyill:
Some small simplifications:

* Remove `describe.only` to run test suite and then fix treatment of `Private` elements.
* Remove scope test. We can assume our SCL file is valid and it's not testable.
* Remove nullish array coalescing, we never need this for these cases.
* Remove `Private` check, we never need this because we have already filtered out `Private` elements.

I'm just getting back to scl-lib after a hiatus so I'm not intending to be especially pedantic, so sorry if this is a bit annoying :crying_cat_face: 

These changes bring our test coverage for `removeIED.ts` up to :100:%

### DIFF
--- a/tIED/removeIED.spec.ts
+++ b/tIED/removeIED.spec.ts
@@ -53,6 +53,12 @@ describe("Function to an remove the IED and its referenced elements", () => {
     expect(numberRemoves(edits, "LNode")).to.equal(1);
   });
 
+  it("returns empty LNode edits for an IED without substation LNodes with preserveLNodes option", () => {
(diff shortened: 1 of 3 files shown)


stee-re:
Wow! When I scanned over the changes, I thought to myself - no way - this is impossible!! But every single thing you found in removeIED was 100% spot on!! Thank you for making all of this better. I'm still embarressed that I let a "describe.only" get committed!!

stee-re (approved):
Thank you for all these improvements - REALLY appriciated!!!

